PR #9, opened 2017-08-23
REPO-2819 : DELETE /shared-links/{id} with admin returns 403

Epurashu:
Add an extra check to QuickShareServiceImpl.canDeleteSharedLink, because the admin user can return the SiteName of a nodeRef (even if it's private) this cause the request to verify if the current user(admin) is a manager/collaborator/someone who shared the link.


montgolfiere:
As per Matt, we should also add/update existing junit test as well (if not already) ?

Epurashu:
For a similar scenario we chose to use TAS instead of Junit because we are jumping between users and event endpoints to accomplish a complex use case( I have updated the TAS test with this scenario), but I will take another look into the Junit and I will add this case.

Epurashu:
Added a JUnit test for this case into this commit: 
https://github.com/Alfresco/alfresco-repository/pull/9/commits/20a6ba2b281d1aa17d59008df35cb593932ed96a
I have added this case to another test that is in the alfresco-remote-api project : 
https://github.com/Alfresco/alfresco-remote-api/commit/43061015e9d3d3e4a3c50d7cfb96e938ced33f0e

PR #14, opened 2017-09-08
Fix/mnt 18404 image magick fails on windows

alandavis:
Contains fix for bug and removal of img.gslib that was not fully removed as part of the REPO-2054.


andrei-rebegea:
other then the comment about "img.gslib" , everything looks great

PR #20, opened 2017-09-19
Fix/mnt 16641 cmis applying aspects including protected properties (REPO-1065)

Epurashu:


src/test/java/org/alfresco/opencmis/CMISTest.java
```diff
@@ -167,6 +167,8 @@
 import org.junit.Test;
 import org.springframework.context.ApplicationContext;
 import org.springframework.extensions.webscripts.GUID;
+
```
derekhulley: This import looks wrong.

src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -3097,13 +3097,13 @@ public void setProperties(NodeRef nodeRef, TypeDefinitionWrapper type, Propertie
                     throw new CmisInvalidArgumentException("Property " + property.getId() + " is unknown!");
                 }
             }
+
```
derekhulley: There should be a single check to see if the node is a working:
`boolean isOnWorkingCopy = checkOutCheckInService.isWorkingCopy(nodeRef);`

src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -3097,13 +3097,13 @@ public void setProperties(NodeRef nodeRef, TypeDefinitionWrapper type, Propertie
...
+            Updatability updatability = propDef.getPropertyDefinition().getUpdatability();
+            if (!isUpdatable(updatability, nodeRef))
```
derekhulley: `
if (!isUpdatable(updatability, isOnWorkingCopy))
`

src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -3579,13 +3559,12 @@ public void setProperty(NodeRef nodeRef, TypeDefinitionWrapper type, String prop
...
         }
-
+
         Updatability updatability = propDef.getPropertyDefinition().getUpdatability();
-        if ((updatability == Updatability.READONLY)
-                || (updatability == Updatability.WHENCHECKEDOUT && !checkOutCheckInService.isWorkingCopy(nodeRef)))
-        {
-            throw new CmisInvalidArgumentException("Property " + propertyId + " is read-only!");
-        }
+        if(!isUpdatable(updatability, nodeRef))
```
derekhulley: `
if (!isUpdatable(updatability, isOnWorkingCopy))
`

src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -4109,5 +4088,18 @@ private CMISRenditionMapping getRenditionMapping()
             singletonCache.put(KEY_CMIS_RENDITION_MAPPING_NODEREF, renditionMapping);
         }
         return renditionMapping;
+    }
+
+    private boolean isUpdatable(Updatability updatability, NodeRef nodeRef)
```
derekhulley: `
private boolean isUpdatable(Updatability updatability, boolean isOnWorkingCopy)
`

src/main/java/org/alfresco/opencmis/CMISConnector.java
derekhulley: The aim is to only check for the working copy once for all properties, reducing a whole bunch of unnecessary work back and forth to check aspects, etc.

src/test/java/org/alfresco/opencmis/CMISTest.java
```diff
@@ -167,6 +167,8 @@
...
+import com.sun.tools.xjc.model.Aspect;
```
derekhulley: Incorrect import.

src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -4109,5 +4089,30 @@ private CMISRenditionMapping getRenditionMapping()
...
+        {
+            return false;
+        }
+        else
+        {
+            return true;
+        }
+    }
+
+    private boolean isUpdatable(Updatability updatability)
```
derekhulley: There is no need to **overload** the _isUpdatable_ method.
Just use Boolean instead of boolean and allow true/false/null.
Also, methods that are inherited will inherit the javadoc as well.
Private methods do not have that luxury: you should add javadoc for the private method to say what it does and how the arguments are used.

derekhulley:
Please merge.

PR #21, opened 2017-09-20
Fix REPO-2901 Service Pack: MNT-18340: Need two CMIS updates to apply metadata and secondary types

mattward:


src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -3213,13 +3213,6 @@ private void processSecondaryTypes(NodeRef nodeRef, List secondaryTypes, Map<Str
         for(QName aspectQName : aspectsToRemove)
         {
             nodeService.removeAspect(nodeRef, aspectQName);
-            // aspect is being removed so remove all of its properties from the propsToAdd map
```
derekhulley: I think this is the actual fix; the line ending changes makes it tricky to see more.

mattward: Apologies, I normally have my IDE set to Windows line endings as per our coding standards, not sure what's happened here - I'll check my settings.

This is indeed the fix, it is to remove the ~half dozen lines that fetches a list of properties and removes them from `propsToAdd`

derekhulley:
After conflicts, please merge.

PR #22, opened 2017-09-22
Fix REPO-2903 Service Pack: MNT-18285: Brute force account attacks: Accounts not reenabled when using authentication chaining

cristianturlica:
- Logins are now protected based on a combined key from authentication service system id and user name, this allows us to fix the case when a valid login was denied for subsequent authentication service if the prior authentication services in the chain failed.
- Cache is now set to be local, as the new implementation doesn't require it to be fully distributed


src/main/java/org/alfresco/repo/security/authentication/AuthenticationServiceImpl.java
```diff
@@ -42,7 +43,10 @@
     private Log logger = LogFactory.getLog(AuthenticationServiceImpl.class);
     AuthenticationComponent authenticationComponent;
     TicketComponent ticketComponent;
-    
+
+    /** a uuid identifying this unique instance */
+    private String uuid;
```
derekhulley: The variable name does not really describe what it does.
Please use something like: `serviceInstanceId`

cristianturlica: I will update it, thanks.

src/main/java/org/alfresco/repo/security/authentication/AuthenticationServiceImpl.java
```diff
@@ -186,10 +195,15 @@ public void recordFailedAuthentication(String userName)
                     }
                 }
             }
-            protectedUsersCache.put(userName, protectedUser);
+            protectedUsersCache.put(protectedUserKey, protectedUser);
         }
     }
 
+    private String getProtectedUserKey(String userName)
```
derekhulley: Javadoc to say what the method does.  It's private so there is no way to guess that it combines the service instance ID with the username.

cristianturlica: I will make it public and add the java doc.

src/main/java/org/alfresco/repo/security/authentication/AuthenticationServiceImpl.java
```diff
@@ -186,10 +195,15 @@ public void recordFailedAuthentication(String userName)
...
+    {
+        return uuid + userName;
```
derekhulley: `return uuid + userName;`
`return uuid + "@@" + userName;`

derekhulley:
None of the commits actually say how the fix was achieved.
Please add it to the merge commit so that we can know what was done to fix the issue without looking at the code.

src/main/java/org/alfresco/repo/security/authentication/AuthenticationServiceImpl.java
derekhulley: Private is fine.  But it needs to be commented.

src/test/java/org/alfresco/repo/security/authentication/AuthenticationServiceImplTest.java
```diff
@@ -57,6 +58,7 @@
     private SimpleCache<String, ProtectedUser> cache;
     private TicketComponent ticketComponent = mock(TicketComponent.class);
     private AuthenticationServiceImpl authService;
+    private AuthenticationServiceImpl authService2;
```
mattward: Would be nice to have a comment why there is a second instance of the ASI class

src/main/java/org/alfresco/repo/security/authentication/AuthenticationServiceImpl.java
cristianturlica: Done

PR #26, opened 2017-09-27
REPO-2926: CMIS update now ignores aspects in the sys namespace (MNT-18340)

mattward:
Fixes CMIS sys:* aspect update behaviour — clients must not manipulate sys aspects via CMIS

This follows on from (and depends upon) the fix for MNT-18340 which has already been merged into develop.

Since this is an API-breaking change, it will go into 6.0+ only (merge to develop with no backport)

*Apologies for the spurious end-of-line changes* — I've now worked out why this is happening. I have my IDE set to remove trailing whitespace upon save, so that I don't save empty lines with the auto-indent left hanging. Unfortunately it seems that a lot of existing code has trailing space that is being manipulated by this feature, so I'll disable it now.

### DIFF
--- a/src/main/java/org/alfresco/opencmis/CMISConnector.java
+++ b/src/main/java/org/alfresco/opencmis/CMISConnector.java
@@ -238,12 +238,14 @@ import org.springframework.context.event.ApplicationContextEvent;
 import org.springframework.extensions.surf.util.AbstractLifecycleBean;
 import org.springframework.util.StringUtils;
 
+import static java.util.Collections.singletonList;
+
 /**
  * Bridge connecting Alfresco and OpenCMIS.
  * <p/>
  * This class  provides many of the typical services that the {@link CmisService} implementation
  * will need to use Alfresco.
- * 
+ *
  * @author florian.mueller
  * @author Derek Hulley
  * @author steveglover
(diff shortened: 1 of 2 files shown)


src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -3183,28 +3185,34 @@ private void processSecondaryTypes(NodeRef nodeRef, List secondaryTypes, Map<Str
...
         Iterator<QName> it = aspectsToRemove.iterator();
         while(it.hasNext())
         {
             QName aspectQName = it.next();
             TypeDefinitionWrapper w = getOpenCMISDictionaryService().findNodeType(aspectQName);
-            if(w == null || secondaryTypeAspects.contains(aspectQName))
+            if(w == null || secondaryTypeAspects.contains(aspectQName) || namespacesToIgnore.contains(aspectQName.getNamespaceURI()))
             {
-                // the type is not exposed or is in the secondary types to set, so remove it from the to remove set
+                // the type is not exposed,
```
derekhulley: Not sure why the comment has to be spread over four lines.

mattward: On one line it goes to column 161, which seems a bit long for my liking

PR #27, opened 2017-09-28
Fix/mnt 18275 bulk

alandavis:
MNT-18275 Change detected mimetype: pdf->ai ps->eps but only if extension is correct

Fix is in alfresco-data-model, but there is also a test for bulk import  in this repository.


alandavis:
PLEASE FIX UP THE pom.xml dependency to alfresco-data-model once that has been released following https://github.com/Alfresco/alfresco-data-model/pull/3.

alandavis:
Cherry picked to develop to avoid conflicts

PR #28, opened 2017-10-05
fix/MNT-18001_ignore_version_label_in_metadata

CodrinChirica:


src/main/java/org/alfresco/repo/bulkimport/metadataloaders/XmlPropertiesFileMetadataLoader.java
```diff
@@ -97,6 +100,15 @@
 {
     private final static Log log = LogFactory.getLog(XmlPropertiesFileMetadataLoader.class);
     private final static String METADATA_FILE_EXTENSION = "properties.xml";
+    // MNT-18001
```
derekhulley: The comments on the private member could instead be added as javadoc comments on the public setter.

derekhulley:
Looks good.
Let's use the squash option.

PR #30, opened 2017-10-09
REPO-2575: added property loader pattern for non-enterprise projects

mattward:
Now that we have "decomposed" various parts of the mega-monolith repository
into smaller projects such as alfresco-remote-api, we need a mechanism for
adding new system properties within the project that requires them.

For example, to add a new boolean/toggle for the WWW-Authenticate behaviour
in the remote API project, we don't want that setting to go in
alfresco-repository - we want it to live in alfresco-remote-api. So this
extra pattern (classpath*:alfresco/project.properties) is being added to
load these project specific properties.


derekhulley:
I think the pattern used will find the first occurence on the classpath, which is not what we want.
Better to use a wildcard in the pattern as well and then each project can use something specific to itself.
`alfresco/project-*.properties`
can be used by remote-api as
`alfresco/project-remote-api.properties`
for example.

mattward:
Thanks Derek, that makes sense - have changed the pattern

PR #31, opened 2017-10-10
Fix/repo 2025 cat tests

alandavis:
REPO-2025 Marked existing tests with categories LuceneTests and DBTests 

Also includes a few PerformanceTests. So that the DB tests could be run on their own for each DB, a new
AllDBTestsTestSuite was created and it was tried out with MySQL (REPO-2002), MariaDB (REPO-2005) and PostgreSQL (REPO-2954).

A few existing TestSuites have been removed because we now have AllDBTestsTestSuite or the same set was covered by one of the TestSuites run in the latest Bamboo build jobs.


alandavis:
OMG it looks like a pull request from hell. However most of it is simply because white space changed when 90 files were marked as LuceneTests.  I have been through all of them with Beyond Compare and that is the only change, so they are okay. Real shame about the white space.

I also check my own changes, but would someone else like to do the same if they can find them. You might need Beyond Compare, or for there to be another commit to fix the white space.

andrei-rebegea:
I reviewed the changes using the ignore whitespace parameter : https://github.com/Alfresco/alfresco-repository/pull/31/files?w=1

PR #33, opened 2017-10-25
REPO-3004: Implement proper cluster behaviour

mmuller88:


src/main/java/org/alfresco/heartbeat/HeartBeat.java
```diff
@@ -120,7 +127,18 @@ public HeartBeat(final ApplicationContext context, final Boolean testMode)
...
 
-            // We force the job to be scheduled regardless of the potential state of the licenses
+            JobLockService jobLockService = null;
+            try
+            {
+                jobLockService = (JobLockService) context.getBean("jobLockService");
+            }
+            catch (NoSuchBeanDefinitionException e)
+            {
+                logger.error("jobLockService not found", e);
```
derekhulley: Is it not possible to inject the jobLockService?
This bean, if not found, is a problem that should prevent the system from starting at all.
As you have it, there will be an error logged followed by NPEs.  Just inject the service you need.

derekhulley:
This commit will work to prevent duplicate work in the cluster.  The behaviour is not correct, yet.
The sender does not need to be job locked.
The collectors need job locks.
The sender schedule must be a configurable frequency.
The collectors must be non-configurable (hard coded in bean xml config) cron expressions.

derekhulley:
Could you review the class comments to reflect what it does in the new world, please?

src/main/resources/alfresco/heartbeat/heartbeat-repo-collector-context.xml
```diff
@@ -13,6 +15,8 @@
 
     <!-- HeartBeat authority data collector -->
     <bean id="authorityDataCollector" class="org.alfresco.heartbeat.AuthoritiesDataCollector" parent="hbBaseDataCollector" init-method="register">
+        <property name="cronExpression" value="0 0/2 * * * ?"/> <!-- every hour -->
+        <constructor-arg type = "java.lang.String" value = "acs.repository.usage.authorities"/>
```
eknizat: Can use the setter instead of constructor to set the collector name
<property name="collectorId" value="acs.repository.usage.authorities">

src/main/resources/alfresco/heartbeat/heartbeat-context.xml
```diff
@@ -16,10 +16,14 @@
     <bean id="hbDataCollectorService" class="org.alfresco.heartbeat.HBDataCollectorServiceImpl">
         <constructor-arg value="${heartbeat.enabled}" />
         <property name="hbDataSenderService" ref="hbDataSenderService" />
+        <property name="scheduler" ref="schedulerFactory" />
+        <property name="jobLockService" ref="jobLockService" />
+        <property name="licenseService" ref="licenseService" />
```
eknizat: Remove <property name="licenseService" ref="licenseService" /> 
This will result in bean not found exception

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -56,44 +83,184 @@
     public HBDataCollectorServiceImpl (boolean defaultHeartBeatState)
     {
         this.defaultHbState = defaultHeartBeatState;
+        licenseService.registerOnLicenseChange(this);
```
eknizat: Remove licenseService.registerOnLicenseChange(this) -  the DescriptorService will register this bean with license service

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -109,4 +276,68 @@ public synchronized void enabled(boolean enabled)
...
+    {
+        public void execute(final JobExecutionContext jobexecutioncontext) throws JobExecutionException
+        {
+            String jobName = jobexecutioncontext.getJobDetail().getName();
+            QName qName = QName.createQName(NamespaceService.SYSTEM_MODEL_1_0_URI, jobName);
+            String lockToken = null;
+            try
+            {
+                // Get a lock
+                lockToken = jobLockService.getLock(qName, LOCK_TTL);
```
derekhulley: If you put a TTL of 30s on the job then you need to be sure that all jobs finish in that time; this cannot be guaranteed.
That is why you should either
- refresh the lock at regular intervals
- use the job lock callback pattern (plenty of examples) and an AtomicBoolean to keep track of whether the job is still active or not.

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -56,44 +83,184 @@
...
+        }
+    }
+
+    private void restartAllCollectorSchedules()
+    {
+        try
+        {
+            for(HBBaseDataCollector collector : collectors)
+            {
+                collectAndSendData(collector);
```
eknizat: scheduleCollector(HBBaseDataCollector collector)  instead of collectAndSendData

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
eknizat: Also set the this.enabled= defaultHeartBeatState

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -44,10 +51,18 @@
...
+    private Scheduler scheduler;
+
+    private boolean testMode = true;
+
+    /** Current enabled state */
+    private boolean enabled = false;
+
     /**
      *
      * @param defaultHeartBeatState the default enabled state of heartbeat
```
eknizat: Also set the enabled state this.enabled = defaultHeartBeatState

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -63,37 +78,158 @@ public void setHbDataSenderService(HBDataSenderService hbDataSenderService)
...
+        boolean newEnabled = !licenseDescriptor.isHeartBeatDisabled();
+
+        if (newEnabled != this.enabled)
         {
-            List<HBData> data = collector.collectData();
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("State change of heartbeat");
+            }
+            this.enabled = newEnabled;
```
eknizat: This class needs to update the hbDataSenderService enabled state whenever the state changes. 
Could call hbDataSenderService.enable(enabled) wherever this.enabled is set or make a method to do both. 
This method used to do that "public synchronized void enabled(boolean enabled)" We don't need the method to be public any more - this was only used by the HeartBeat class

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -63,37 +78,158 @@ public void setHbDataSenderService(HBDataSenderService hbDataSenderService)
...
+            {
+                if (logger.isDebugEnabled())
+                {
+                    logger.debug("Skipping scheduling because of invalid cron expression: " + cronExpression);
+                }
             }
-            catch (Exception e)
+            if(cronTrigger != null)
+            {
+                scheduler.scheduleJob(jobDetail, cronTrigger);
```
eknizat: Could this be moved inside of the try block?

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -63,37 +78,158 @@ public void setHbDataSenderService(HBDataSenderService hbDataSenderService)
...
+            }
+        }
+        else
+        {
+            if (logger.isDebugEnabled())
             {
-                // Log exception
-                logger.error(e);
+                logger.debug("heartbeat job unscheduled");
```
eknizat: I think our logger messages should be updated to reflect which collector is being scheduled/unscheduled. Like this we will just see "heartbeat job unscheduled" multiple times in the log

ancutam:
Reminder: Please squash commits when doing the merge and add a meaningful message to the commit.

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -225,24 +226,31 @@ private synchronized void scheduleCollector(HBBaseDataCollector collector) throw
...
     }
 
+    private void unscheduleJob(String triggerName, HBBaseDataCollector collector) throws SchedulerException
+    {
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("heartbeat job unscheduled with collector: " + collector.getCollectorId() + " " + collector.getCollectorVersion());
+        }
+        scheduler.unscheduleJob(triggerName, Scheduler.DEFAULT_GROUP);
+    }
```
eknizat: Do the logging after calling scheduler.unscheduleJob(triggerName, Scheduler.DEFAULT_GROUP); or change the message. As it is now it will log that heart beat job was unscheduled before actually unscheduling so if this throws an exception the log will say that the job is unscheduled when in fact it isn't

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -56,57 +71,186 @@
...
+        {
+            for(HBBaseDataCollector collector : collectors)
+            {
+                scheduleCollector(collector);
+            }
+        }
+        catch (Exception e)
+        {
+            logger.error("Unable to schedule heart beat", e);
+        }
```
eknizat: Move the try catch inside of the for loop and log exactly which collector failed to schedule/unschedule its job. 
As it is now we will break out of the for loop at the first iteration that throws an exception.
The behaviour we probably want is to log when a collector fails to schedule/unschedule job and to continue the for loop for the other collectors.
Would we want some kind of transaction for this?

src/main/java/org/alfresco/heartbeat/HeartBeatJob.java
```diff
@@ -0,0 +1,198 @@
...
+        try
+        {
+            List<HBData> data = collector.collectData();
+            try
+            {
+                hbDataSenderService.sendData(data);
+            }
+            catch (Exception e)
+            {
+                logger.error(e);
```
eknizat: Is there a reason why we are logging the exception from the sender as error but as debug/warn when it is from the collector?

eknizat: Just checked and saw that that is how it was done before..

src/main/java/org/alfresco/heartbeat/HBBaseDataCollector.java
```diff
@@ -36,10 +36,41 @@
   *
  * @author eknizat
  */
-public abstract class HBBaseDataCollector
-{
+public abstract class HBBaseDataCollector {
```
ancutam: formatting please

src/test/java/org/alfresco/heartbeat/HBDataCollectorServiceImplTest.java
```diff
@@ -52,76 +61,168 @@
...
+
+        dataCollectorService.registerCollector(collector2);
+
+        verifyNumbersOfUnschedulesAndSchedules(scheduler, 2, 1);
+    }
 
     @Test
-    public void testHBDataSenderServiceEnabledChange()
+    public void testSchedulingWithDefaultEnabled() throws Exception
     {
```
eknizat: It looks like you covered the new functionality here which is good. However, maybe it is just a matter of taste but you could split this test up into its respective components. For example, if you have a look at my old HeartBeatTest you will see that each scenario is in its own test method. That way if one of the scenarios fails we will know what is going wrong and it is easy to see that each scenario is covered. The new HBDataCollectorServiceImpl should pass all of the old tests plus more.

src/test/java/org/alfresco/heartbeat/HBDataCollectorServiceImplTest.java
```diff
@@ -52,76 +61,168 @@
...
+        Scheduler scheduler = spy(StdSchedulerFactory.getDefaultScheduler());
+        dataCollectorService.setScheduler(scheduler);
+
+        // try with correct cron expression
+        SimpleHBDataCollector collector1 = new SimpleHBDataCollector("collector1");
+        collector1.setCronExpression("0 0 0/1 * * ?");
+
+        dataCollectorService.registerCollector(collector1);
+
+        verifyNumbersOfUnschedulesAndSchedules(scheduler, 1, 1);
```
eknizat: Could verify that the job is scheduled with the expected job name and that it actually is in the scheduler

src/test/java/org/alfresco/heartbeat/HBDataCollectorServiceImplTest.java
eknizat: Just for reference, here are the scenarios covered previously :
* test that collectors get scheduled when HB is enabled by default - not through license
* test that collectors **don't** get scheduled when HB is disabled by default - not through license
* test that collectors are not scheduled when HB is disabled by default **but** all are scheduled once enabled in license
* test that collectors are scheduled when HB is enabled by default **but** all are unscheduled once disabled in license
* test that license fail reverts back to default enabled/disabled state and all collectors are scheduled/unscheduled

The one difference is that before the tests checked a single HB job and now we need to check jobs for all collectors

PR #34, opened 2017-10-31
REPO-2028: remove lucene dependency 

andrei-rebegea:


killerboot:
@derekhulley 
Please pay attention to org.alfresco.repo.transfer.RepoTransferReceiverImpl#getTempFolder changes. The search query was changed to a translatable to DB query.

PR #36, opened 2017-11-06
REPO-3013 Fix security config for PermissionService

killerboot:
The PermissionService had security config for two methods missing:
* hasReadPermission
* getAuthorisations

getAuthorisations was duplicated as ACL_DENY as it is a system only performance method.


derekhulley:
Changes look good.

PR #37, opened 2017-11-09
REPO-3029: MNT-13055 execute rules subfolders

mattward:
When executing rules "manually" in Share the request to execute on subfolders was being ignored if the folder was set to "do not inherit rules". The recursion was relying on inheritance of rules, but really the two are separate functionality.

It even turned out that the test `testExecution()` was proving this didn't work: it executed the rules with the recursion option but then tested that the aspect _hadn't_ been added to the items in the subfolder. In the final assertion block:

```
assertFalse(nodeService.hasAspect(doc3, ContentModel.ASPECT_VERSIONABLE));
assertFalse(nodeService.hasAspect(doc3, ContentModel.ASPECT_CLASSIFIABLE));
assertTrue(nodeService.hasAspect(doc3, ContentModel.ASPECT_TITLED));
```

`doc3` should have been affected by all 3 reals due to the recursion option, yet we were only testing for the third rule applied directly to that folder.


derekhulley:
Very thorough.  Looks good.

PR #39, opened 2017-11-13
REPO-3011: MNT-17841: The log message 'Full synchronization with user…

ramunteanu:
… registry' is logged when doing a differential sync

   - Removed incorrect warning log message.

### DIFF
--- a/src/main/java/org/alfresco/repo/security/sync/ChainingUserRegistrySynchronizer.java
+++ b/src/main/java/org/alfresco/repo/security/sync/ChainingUserRegistrySynchronizer.java
@@ -714,9 +714,6 @@ public class ChainingUserRegistrySynchronizer extends AbstractLifecycleBean
                     }
                     if (isFullSync && ChainingUserRegistrySynchronizer.logger.isWarnEnabled())
                     {
-                        ChainingUserRegistrySynchronizer.logger
-                                .warn("Full synchronization with user registry '"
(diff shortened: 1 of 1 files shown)


derekhulley:
looks good.  Pls merge.

src/main/java/org/alfresco/repo/security/sync/ChainingUserRegistrySynchronizer.java
```diff
@@ -714,9 +714,6 @@ public Object execute() throws Throwable
                     }
                     if (isFullSync && ChainingUserRegistrySynchronizer.logger.isWarnEnabled())
                     {
-                        ChainingUserRegistrySynchronizer.logger
-                                .warn("Full synchronization with user registry '"
```
mattward: This does at least remove the confusing logging, which is good. But does this mean the variable name is just plain wrong? `isFullSync == true` doesn't mean that a full sync is happening?

ramunteanu: Yes, that's what got me confused and described in this comment https://issues.alfresco.com/jira/browse/MNT-17841?focusedCommentId=543902&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-543902. Thank you for reviewing.

PR #41, opened 2017-11-14
REPO-3040 Make TransferVersionCheckerImpl less restrictive

killerboot:
As FTR is now a separate project it is not feasible to check the minor version and edition during transfer.
Now only major version is checked (6 in this case).


alandavis:
Looks good, following the additional check that the from major version is not null.

PR #42, opened 2017-11-16
MNT-17919 Slow query during acl tracking

andyhind:
Modified ACL count query
Added indexes to support queries by time and id doe acl change set

No fixes required for MySQL

It is possible that some index duplication could be avoided for DBs that can use leading columns for enforcing FK constraints. This can be addressed later and is not likely to be a significant issue.


cristianturlica:
Changes look good to me.

PR #44, opened 2017-11-20
REPO-3034 / MNT-17162: thumbnail.get webscript throws error on delete

ancutam:
The GET thumbnail webscript should not attempt to delete empty thumbnails, since the underlying transaction is marked as read-only.  
Instead, the "corrupted" thumbnail removal should be done in `ThumbnailService`, after the transaction in which the thumbnail node is retrieved, commits, in a post-transaction callback.

The full fix will also require deleting the code that removes the thumbnail from `thumbnail.get.js` in alfresco-remote-api
```
// Remove broken thumbnail
if (thumbnail != null)
{
    thumbnail.remove();
}
```
Note to the fix: Once the corrupted, empty thumbnail is deleted, the only way to recreate it is to explicitly call the get thumbnail webscript. A refresh on the node in the document library will not trigger it. Resetting/deleting the value for `lastModified` might be required.


aforascu:
I have added code to update "lastThumbnailModification" before deleting the thumbnail so that a refresh in Share will regenerate it.

PR #45, opened 2017-11-20
REPO-2627: HB2.0 collectors, schedules and clustering

eknizat:
* Add support for individual collector schedules via cron expressions
* Add job locking to prevent duplicate collector work in cluster


pom.xml
```diff
@@ -2,7 +2,7 @@
     <modelVersion>4.0.0</modelVersion>
     <artifactId>alfresco-repository</artifactId>
     <name>Alfresco Repository</name>
-    <version>6.12-SNAPSHOT</version>
+    <version>6.12-HeartBeat-1-SNAPSHOT</version>
```
skopf: We need to reset this before merging to master.

src/main/java/org/alfresco/heartbeat/HBBaseDataCollector.java
```diff
@@ -36,12 +36,48 @@
...
+    private String collectorId;
+    private String collectorVersion;
+    private String cronExpression;
+
+    /**
+     * The collector service managing this collector.
+     */
     private HBDataCollectorService hbDataCollectorService;
 
+    public HBBaseDataCollector(String collectorId)
```
skopf: We need both, the collectorId and the version. Why do we pass one as constructor argument and the other one with a setter?

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -44,69 +48,222 @@
...
-                logger.error(e);
+                // Log and ignore
+                logger.error("HeartBeat failed to restart collector: " + collector.getCollectorId() ,e);
             }
         }
     }
 
-    @Override
-    public boolean isEnabledByDefault()
+    private void enable(boolean enable)
```
skopf: As a setter, shouldn't this be named "setEnabled" ?

src/main/java/org/alfresco/heartbeat/HeartBeatJob.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /** Time to live 5 seconds */
+    private static final long LOCK_TTL = 5000L;
+    /** Additional 5 seconds for how much longer the log will be kept */
+    private static final long LOCK_TTL_OFFSET = 5000L;
+
+    public void execute(final JobExecutionContext jobexecutioncontext) throws JobExecutionException
+    {
+        final JobDataMap dataMap = jobexecutioncontext.getJobDetail().getJobDataMap();
+        final HBBaseDataCollector collector = (HBBaseDataCollector) dataMap.get("collector");
```
skopf: The keys in the Map ("collector", "hbDataSenderService" and "jobLockService") should be defined as ```public static final String``` fields on this class, and then referenced from the HBDataCollectorServiceImpl.java. Otherwise, it is easy to forget one or introduce a typo when changing names.

src/main/java/org/alfresco/heartbeat/HBDataCollectorServiceImpl.java
```diff
@@ -44,69 +48,222 @@
...
+            unscheduleJob(triggerName, collector);
+        }
+
+    }
+
+    private void scheduleJob(final String jobName, final String triggerName, final HBBaseDataCollector collector) throws ParseException, SchedulerException
+    {
+        final JobDetail jobDetail = new JobDetail(jobName, Scheduler.DEFAULT_GROUP, HeartBeatJob.class);
+        final String cronExpression = testMode ? testCronExpression : collector.getCronExpression();
+        jobDetail.getJobDataMap().put("collector", collector);
```
skopf: The keys in the Map ("collector", "hbDataSenderService" and "jobLockService") should be defined as ```public static final String``` fields on the HeartBeatJob class, and then referenced from the HBDataCollectorServiceImpl.java. Otherwise, it is easy to forget one or introduce a typo when changing names.

src/main/java/org/alfresco/heartbeat/HeartBeatJob.java
```diff
@@ -0,0 +1,209 @@
...
+            if (logger.isDebugEnabled())
+            {
+                // Verbose logging
+                logger.debug("Heartbeat job failure from collector: " + collector.getCollectorId(), e);
+            }
+            else
+            {
+                // Heartbeat errors are non-fatal and will show as single line warnings
+                logger.warn(e.toString());
+                throw new JobExecutionException(e);
```
skopf: It should not depend on the log level whether we throw an exception or not.
Even if debig is enabled, it ios still fine to log it out as a warn. And we should either always or never throw a JobExecutionException in this situation.

src/main/java/org/alfresco/heartbeat/HBBaseDataCollector.java
eknizat: I thought about this one as well, wanted to see if anyone will spot it.

src/main/java/org/alfresco/heartbeat/HeartBeatJob.java
eknizat: Well spotted, Stefan.

PR #56, opened 2017-12-21
REPO-1886 / ALF-21822: NodeServiceXPath: Unsupported operation list.clear()

aforascu:
Fix for REPO-1886 / ALF-21822: create new list instead of using the old list. 
For some reason it seems that the retrieved list may be used in other operations, so instead of mutating this one to remove duplicates, a new list is created and returned without duplicates. Couldn't reproduce the issue using acp import as suggested in ALF-21822

### DIFF
--- a/src/main/java/org/alfresco/repo/search/NodeServiceXPath.java
+++ b/src/main/java/org/alfresco/repo/search/NodeServiceXPath.java
@@ -195,7 +195,7 @@ public class NodeServiceXPath extends BaseXPath
         Set<Object> set = new HashSet<Object>(resultsWithDuplicates);
         
         // return new list without duplicates
-        List<Object> results = new ArrayList<Object>();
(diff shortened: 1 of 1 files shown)


src/main/java/org/alfresco/repo/search/NodeServiceXPath.java
```diff
@@ -194,9 +194,8 @@ public List selectNodes(Object arg0) throws JaxenException
         
         Set<Object> set = new HashSet<Object>(resultsWithDuplicates);
         
-        // now return as a list again
-        List<Object> results = resultsWithDuplicates;
-        results.clear();
+        // return new list without duplicates
+        List<Object> results = new ArrayList<Object>();
```
ancutam: You can remove the Object type argument:  
`results = new ArrayList<>();`

ancutam:
Looks good, please merge.

PR #57, opened 2017-12-21
REPO-3111: Community: ALF-21962: Webscripts - node.qnamePath incorrect

andrei-rebegea:
ALF-21962: Webscripts - node.qnamePath incorrectly returns : for items without namespace prefix


ramunteanu:
Test looks good to me, can be merged.

PR #58, opened 2018-01-04
REPO-2957: Community: ALF-21948 Bug with setting CC and BCC fields

mihamimi:
Set correct values for bcc adress and fix ClassCastException.


andrei-rebegea:
looks good for merge to me

PR #59, opened 2018-01-05
REPO-2918: ALF-21947 EmailUserNotifier doesn't regard recipient locale preferences

andrei-rebegea:
REPO-2918: ALF-21947 EmailUserNotifier doesn't regard recipient locale preferences
MNT-15287: Activiti Workflow Email subject for workflow are in English while the content is are French

### DIFF
--- a/src/main/java/org/alfresco/repo/notification/EMailNotificationProvider.java
+++ b/src/main/java/org/alfresco/repo/notification/EMailNotificationProvider.java
@@ -28,6 +28,7 @@ package org.alfresco.repo.notification;
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import org.alfresco.error.AlfrescoRuntimeException;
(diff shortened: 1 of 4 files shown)


src/main/java/org/alfresco/repo/notification/EMailNotificationProvider.java
```diff
@@ -209,7 +210,17 @@ public void sendNotification(NotificationContext notificationContext)
         {
             mail.setParameterValue(MailActionExecuter.PARAM_SUBJECT, subject);
         }
-        
+
+        // Note that the smart logic in MailActionExecuter should recognize any
+        // preferred locale for the users in the recipient field and override this value
+        // Note that the smart logic in MailActionExecuter may be broken if we send email addresses
+        // instead of userIDs
+        Locale userLocale = I18NUtil.getLocale();
```
killerboot: Are you sure that the correct locale will be set in this thread?

killerboot:
I don't think there should be any problems in creating unit tests for org.alfresco.repo.activities.feed.EmailUserNotifier#notifyUser and org.alfresco.repo.notification.EMailNotificationProvider#sendNotification.

Mock the necessary services and call the methods directly. Check that the Action object passed in the ActionService in the end has the required locale set as a parameter.
The interesting bit is how you obtain the locale. org.alfresco.util.EmailHelper#getUserLocaleOrDefault seems valid, I'm not so sure about org.springframework.extensions.surf.util.I18NUtil#getLocale.

src/main/java/org/alfresco/repo/notification/EMailNotificationProvider.java
andrei-rebegea: Yes, the browser preferences should be available in this thread that handles the request.

andrei-rebegea:
Alex, I see very little value in adding junit tests to org.alfresco.repo.activities.feed.EmailUserNotifier#notifyUser and org.alfresco.repo.notification.EMailNotificationProvider#sendNotification.
At this point, adding the PARAM_LOCALE parameter is done consistently to all requests, not depending on complex logic. EmailHelper.getUserLocaleOrDefault  is tested separately in another junit. Parameter passing to Action service should be a problem (not something that the email tests should cover). User preferences are/should be tested somewhere else. Therefore I feel that the effort of adding these 2 new unit tests is not worth it - I don't see the value.

This fix will not fix all the problem related to localization in the emailing department (see REPO-807 ), just these specific ones.

killerboot:
Makes sense. Thanks and please merge.

PR #61, opened 2018-01-09
REPO-1250: Community: ALF-21750: FileNameValidator.getValidFileName a…

mihamimi:
…llows a dot in the end of a filename

   Added a validation check for dot at the end of the filename. The dot will be converted into underscore.

### DIFF
--- a/src/main/java/org/alfresco/util/FileNameValidator.java
+++ b/src/main/java/org/alfresco/util/FileNameValidator.java
@@ -50,11 +50,11 @@ public class FileNameValidator
 
     public static boolean isValid(String name)
     {
-        if (!isValidRegex(name) || name.endsWith("."))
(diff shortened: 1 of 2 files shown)


src/main/java/org/alfresco/util/FileNameValidator.java
```diff
@@ -42,12 +42,21 @@
...
+
+    private static boolean isValidRegex(String name)
     {
         return !FILENAME_ILLEGAL_PATTERN_REPLACE.matcher(name).find();
     }
-    
+
+    public static boolean isValid(String name)
+    {
+        if (!isValidRegex(name) || name.endsWith("."))
```
killerboot: I would change this to one line:
!name.endsWith(".") && isValidRegex(name)

PR #63, opened 2018-01-10
REPO-2182 : Community: ALF-21852: setContentStream() returns the wrong document version when auto-versioning is turned on

Epurashu:
Separete into different new transactions writing content from creatingObjectId for setContentStream()
   Separate into different new transactions deleteing content from creatingObjectId for deleteContentStream()
   Separate into different new transations appending content from creatingObjectId for appendContentStream()
   Create JUnit CMISTest.testSetDeleteAppendContentStreamVersioning() testing that set/delete/apped calls return the new version.

### DIFF
--- a/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
+++ b/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
@@ -65,6 +65,7 @@ import org.alfresco.repo.node.getchildren.GetChildrenCannedQuery;
 import org.alfresco.repo.security.authentication.AuthenticationUtil;
 import org.alfresco.repo.security.authentication.Authorization;
 import org.alfresco.repo.transaction.RetryingTransactionHelper;
+import org.alfresco.repo.transaction.RetryingTransactionHelper.RetryingTransactionCallback;
 import org.alfresco.repo.version.VersionModel;
 import org.alfresco.service.cmr.model.FileInfo;
 import org.alfresco.service.cmr.model.FileNotFoundException;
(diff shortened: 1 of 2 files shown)


src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
```diff
@@ -1526,15 +1527,32 @@ public void appendContentStream(String repositoryId, Holder<String> objectId, Ho
...
+               }
+                return null;
+            }}, false, true);
+
+       String objId = helper.doInTransaction(new RetryingTransactionCallback<String>()
+       {
+           public String execute() throws Throwable
+           {
+               return connector.createObjectId(nodeRef);
+       }}, false, true);
```
mmuller88: formation mistake?

src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
```diff
@@ -1569,25 +1587,41 @@ public void setContentStream(
...
+                return null;
+            }}, false, true);
 
-        objectId.setValue(connector.createObjectId(nodeRef));
+        String objId = helper.doInTransaction(new RetryingTransactionCallback<String>()
+        {
+            public String execute() throws Throwable
+            {
+                return connector.createObjectId(nodeRef);
+        }}, false, true);
```
mmuller88: here too

src/test/java/org/alfresco/opencmis/CMISTest.java
```diff
@@ -3255,6 +3256,89 @@ private void checkDefs(List<TypeDefinitionWrapper> defs)
         withCmisService(callback, CmisVersion.CMIS_1_1);
     }
 
+    /**
+     * Test auto version behavior for setContentStream, deleteContentStream and appendContentStream according to REPO-2182.
```
mmuller88: Does it not has a Alf or MNT number

derekhulley:
In some respects, this is a workaround for the fact that the versioning behaviour is only processed at the end of the transaction (I think that's what we're looking to work around in your fix).

src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
```diff
@@ -1526,15 +1527,32 @@ public void appendContentStream(String repositoryId, Holder<String> objectId, Ho
...
+                   connector.appendContent(info, contentStream, isLastChunk);
+               }
+               catch(IOException e)
+               {
+                    throw new ContentIOException("", e);
+               }
+                return null;
+            }}, false, true);
+
+       String objId = helper.doInTransaction(new RetryingTransactionCallback<String>()
```
derekhulley: This second transaction is independent of the first inner transaction; we've lost atomicity of the overall call.

src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
derekhulley: Again, atomicity is broken unless the second inner transaction is read-only.

derekhulley:
That looks functional now.
Of course, there are two extra transactions per call, including the now-redundant transaction that wraps the method call.

An alternative approach would be to register a transaction listener to update the object ID is a post-transaction callback (afterCommit).  You would still need one of the read-only transactions but in a different phase.

On balance, I think the code you have is clearer than the alternative.

PR #64, opened 2018-01-11
REPO-2253/ALF-21854 Fix action parameter locale lookup

mattward:
REPO-2253: Community: ALF-21854 Action parameter lookup for "de_DE" falls back to "root" locale instead of "de"

This fixes the locale fallback problem for action definitions, such that if a parameter definition exists for German ("de") and the client asks for it with the locale Germany ("de-DE") then the labels will be returned in German, rather than English, even though there is no specific language bundle for the de-DE locale.


derekhulley:
Makes sense.  I'm glad there was a method to get the nearest match.

ramunteanu:
Changes look good to me, can be merged.

PR #66, opened 2018-01-12
REPO-3176: Migrate relevant community architecture map pages to Repository source code repo

dcaruana:
In this request you'll find...

1) A consolidated index page listing all the community repository components with links to documentation pages (where available)
2) Relevant documentation pages from architecture map
3) Version badges have been removed as the versions do not relate to community releases

NOTE: the public api is not part of the code repository. It needs to be moved to the remote-api source code repo which will be done in a separate ticket at a later date.


bremmington:
I think we should remove the "Identity" and "Queue". We should talk about how/where we want to document future things generally, but these are really a platform services things.

PR #72, opened 2018-02-01
Feature/repo 1066 tika upgrade

andrei-rebegea:


andrei-rebegea:
The related TikaMetadataKeys.TIKA_PARSER_PARSE_SHAPES_KEY will be handled in https://issues.alfresco.com/jira/browse/REPO-3251 as a decision needs to be taken by the PO

src/main/java/org/alfresco/repo/content/metadata/MetadataExtracterConfigImpl.java
```diff
@@ -53,7 +53,7 @@ public void prepareMetadataWithConfigParams(Metadata metadata)
             return;
         }
         boolean shouldParseShapes = getBooleanProperty(PARSE_SHAPE_PROP_STRING, TIKA_PARSER_PARSE_SHAPES_DEFAULT_VALUE);
-        metadata.add(TikaMetadataKeys.TIKA_PARSER_PARSE_SHAPES_KEY, Boolean.toString(shouldParseShapes));
+       // metadata.add(TikaMetadataKeys.TIKA_PARSER_PARSE_SHAPES_KEY, Boolean.toString(shouldParseShapes));
```
killerboot: Maybe this should be removed at all.

src/main/java/org/alfresco/repo/content/transform/TextToPdfContentTransformer.java
```diff
@@ -1,56 +1,46 @@
...
-
 import org.alfresco.error.AlfrescoRuntimeException;
 import org.alfresco.repo.content.MimetypeMap;
-import org.alfresco.service.cmr.repository.ContentReader;
-import org.alfresco.service.cmr.repository.ContentWriter;
-import org.alfresco.service.cmr.repository.TransformationOptionLimits;
-import org.alfresco.service.cmr.repository.TransformationOptionPair;
-import org.alfresco.service.cmr.repository.TransformationOptionPair.Action;
-import org.alfresco.service.cmr.repository.TransformationOptions;
+import org.alfresco.service.cmr.repository.*;
```
killerboot: I don't think such imports are allowed by code formatting standards.

src/main/java/org/alfresco/util/json/AlfrescoJsonUtils.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+package org.alfresco.util.json;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class AlfrescoJsonUtils
```
killerboot: Was this class created just for 1 usage in PreferenceServiceImpl?

andrei-rebegea: It is also used in some tests in remote-api project. it would be deleted probably when we remove that json lib from our dependencies

PR #73, opened 2018-02-02
MNT-19061: Title property is garbled when uploading a Japanese HTML d…

cristianturlica:
…ocument

   - updated extracter to use the charset found in file


CodrinChirica:
Changes look good to me, it can be merged.

PR #78, opened 2018-02-14
REPO-3255: Service Pack: MNT-19181 and MNT-19129: Authentication regressions

andrei-rebegea:
ignore authentication protection for ext auth, and use the unsercure personService bean to be able to check if the system uses case sensitive usernames


killerboot:
LGTM, it might be worth investing some time in unit tests.

PR #79, opened 2018-02-16
Repo 3195 trial license

alandavis:
REPO-3195 Heartbeat Addition of deregister method so that a collector can remove itself once it is done.


alandavis:
Checked by Martin

PR #87, opened 2018-03-05
REPO-3351: EOL Lucene: Remove NodeIndexer and index 

derekhulley:
Remove NodeIndexer
Remove Lucene searcher-related code

Required as a step towards REPO-1324: EOL lucene-*;  Unfortunately, the parsing means that we cannot drop the jars.  However, we can remove everything that creates with Lucene indexing constructs.


src/main/java/org/alfresco/repo/admin/ConfigurationChecker.java
```diff
@@ -101,7 +90,6 @@ public String toString()
     {
         StringBuilder sb = new StringBuilder(50);
         sb.append("ConfigurationChecker")
-          .append("[indexRecoveryMode=").append(indexRecoveryMode)
           .append("]");
         return sb.toString();
```
eknizat: No need for this "]" now

PR #99, opened 2018-03-26
Implement CQRS event source, service, ibatis cars writer, some simple…

mmuller88:
… cqrs service tests.


derekhulley:
Interesting bit to play with.
Consider that the existence check for the node id does not make sense since the node ID is generated by the database.  The existence check needs to be the NodeRef; ultimately the database is the arbiter of uniqueness in this respect.

I would encourage a play with things at a higher level than individual inserts.  Have a chat and we can point you at places where the pattern should be applied.

PR #102, opened 2018-03-29
MNT-16433: Add ftp.externalAddress property

alxgomz:
Adds a property to collect the value to pass on to JLAN to use as the
FTPExternalAddress parameter.

> This PR is for MNT-16433 which first requires jlan patch. 
the pom file need to reference a jlan artifact which has the corresponding code. Please see https://git.alfresco.com/Repository/jlan/merge_requests/3


derekhulley:
Issue hs been resolved as fixed; although this PR was not taken directly, the change have been incorporated.

PR #104, opened 2018-04-02
REPO-3280: Content Model admin security bypass

anechifor:
Validate location when changing type to "cm:dictionaryModel".

### DIFF
--- a/src/main/java/org/alfresco/repo/dictionary/DictionaryModelType.java
+++ b/src/main/java/org/alfresco/repo/dictionary/DictionaryModelType.java
@@ -79,7 +79,8 @@ public class DictionaryModelType implements ContentServicePolicies.OnContentUpda
                                             NodeServicePolicies.OnDeleteNodePolicy,
                                             NodeServicePolicies.OnCreateNodePolicy,
                                             NodeServicePolicies.OnRemoveAspectPolicy,
-                                            NodeServicePolicies.OnSetNodeTypePolicy
+                                            NodeServicePolicies.OnSetNodeTypePolicy,
+                                            NodeServicePolicies.BeforeCreateNodePolicy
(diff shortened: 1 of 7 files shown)


src/main/java/org/alfresco/repo/dictionary/DictionaryModelType.java
```diff
@@ -434,7 +475,60 @@ public void onCreateNode(ChildAssociationRef childAssocRef)
...
+    private boolean isValidLocation(NodeRef definitionNode)
+    {
+        StoreRef storeRef = repositoryModelsLocation.getStoreRef();
+        NodeRef rootNode = nodeService.getRootNode(storeRef);
+        List<NodeRef> nodeRefs = searchService.selectNodes(rootNode, repositoryModelsLocation.getPath(), null, namespaceService, false);
+
+        if (nodeRefs.isEmpty() || nodeRefs.size() > 1)
+        {
+            throw new IllegalStateException(
+                    "Incorrect number of nodes (" + nodeRefs.size() + ")" + " found for workflow location: " + repositoryModelsLocation.getPath());
```
skopf: copy/paste typo:
It is "models location" instead of "workflow location"

src/main/java/org/alfresco/repo/dictionary/DictionaryModelType.java
```diff
@@ -434,7 +475,60 @@ public void onCreateNode(ChildAssociationRef childAssocRef)
...
+        }
+
+        NodeRef modelParent = nodeRefs.get(0);
```
skopf: We have the same code here as in "isValidLocation()". We could extract this code into a private method "private NodeRef getModelsLocation()" or so.

PR #106, opened 2018-04-09
REPO-3447: Remove c3p0

killerboot:
Quartz uses c3p0 jdbc pool for it's distributed job storage since version 2.x
This commit will exclude c3p0 lib as distributed job storage is not used.


derekhulley:
Let's go with it.

PR #108, opened 2018-04-13
REPO-3329 record deployment method

eknizat:
* Add DeploymentMethodProvider which reads the deployment method property "deployment.method"
* Extend InfoDataCollector to include deployent method value
* Set default value for deployment.method property in repository.properties

### DIFF
--- a/src/main/java/org/alfresco/heartbeat/InfoDataCollector.java
+++ b/src/main/java/org/alfresco/heartbeat/InfoDataCollector.java
@@ -26,6 +26,7 @@
 package org.alfresco.heartbeat;
 
 import org.alfresco.heartbeat.datasender.HBData;
+import org.alfresco.repo.deployment.DeploymentMethodProvider;
 import org.alfresco.repo.descriptor.DescriptorDAO;
 import org.alfresco.service.descriptor.Descriptor;
 import org.alfresco.util.PropertyCheck;
@@ -54,6 +55,7 @@ import java.util.*;
  *          </li>
  *          <li><b>schema:</b> Int - The schema number. {@link Descriptor#getSchema()}</li>
  *          <li><b>edition:</b> String - The edition. {@link Descriptor#getEdition()}</li>
+ *          <li><b>deploymentMethod:</b> String - The deployment method used to deploy this Alfresco instance. {@link DeploymentMethodProvider#getDeploymentMethod()}</li>
  *      </ul>
  *  </li>
  * </ul>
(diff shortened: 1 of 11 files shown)


src/main/java/org/alfresco/heartbeat/InfoDataCollector.java
```diff
@@ -115,6 +125,10 @@ public void afterPropertiesSet() throws Exception
         infoValues.put("version", version);
         infoValues.put("schema", new Integer(serverDescriptor.getSchema()));
         infoValues.put("edition", serverDescriptor.getEdition());
+        infoValues.put("deploymentMethod", deploymentMethodProvider.getDeploymentMethod().toString());
+
+        // remove this
```
derekhulley: I agree.

src/main/resources/alfresco/heartbeat/heartbeat-context.xml
```diff
@@ -20,6 +20,7 @@
         <property name="hbDataSenderService" ref="hbDataSenderService" />
         <property name="scheduler" ref="schedulerFactory" />
         <property name="jobLockService" ref="jobLockService" />
+        <property name="testMode" value="true" />
```
derekhulley: Is this correct?  Should this be in test mode by default?

src/main/resources/alfresco/core-services-context.xml
```diff
@@ -1352,4 +1352,11 @@
       <property name="policyComponent" ref="policyComponent" />
       <property name="namespaceService" ref="namespaceService" />
    </bean>
+
+   <bean id="deploymentMethodProvider" class="org.alfresco.repo.deployment.DeploymentMethodProviderImpl" >
+      <property name="deploymentMethodAsString">
```
derekhulley: I think that Spring will do the conversion.  Have you checked?

src/main/resources/alfresco/heartbeat/heartbeat-context.xml
eknizat: This is just to schedule all collectors at 1min intervals, I will remove it before merge

src/main/resources/alfresco/core-services-context.xml
eknizat: Yes you are right, Spring does indeed do the conversion. I'll change it in the next commit.

derekhulley:
Could you add a comment on why this MR was closed, please @eknizat.

eknizat:
The MR was closed and its branch deleted and replaced by this MR: https://github.com/Alfresco/alfresco-repository/pull/111
This was done for clarity as the original branch was used to investigate a failing test and had irrelevant commits to the REPO-3329 feature.

PR #111, opened 2018-04-18
REPO-3329: Record deployment method

eknizat:
* Add DeploymentMethodProvider which reads the deployment method property
* Set default value for deployment method in repository.properties
* Add deployment method to InfoDataCollector


killerboot:
Please extract the test fix in NodeCrawlerTest in a separate commit on master.
Otherwise LGTM.

PR #113, opened 2018-04-27
REPO-3465  Add CSRF filter config

killerboot:
CSRF filter is now configured by these props:
csrf.filter.referer=
csrf.filter.referer.always=false
csrf.filter.origin=
csrf.filter.origin.always=false

Referer and origin here is a regexp and should look like:
http://mydomain:80/*.


alandavis:
Looks fine - thanks for the walk through.

eknizat:
LGTM

PR #116, opened 2018-05-07
MNT-19474: update schema method should check that all patches are applied

andrei-rebegea:


andrei-rebegea:
I investigated the logs from the kubernetes failed startups
{code}
2018-05-02 12:01:59,816  INFO  [domain.schema.SchemaBootstrap] [localhost-startStop-1] Schema managed by database dialect org.alfresco.repo.domain.dialect.PostgreSQLDialect.
2018-05-02 12:02:08,505  INFO  [domain.schema.SchemaBootstrap] [localhost-startStop-1] No changes were made to the schema.
2018-05-02 12:02:08,526  INFO  [domain.schema.SchemaBootstrap] [localhost-startStop-1] Another cluster node is doing a DB schema initialization or DB update. Waiting for the other node to finish...
2018-05-02 12:02:08,527  ERROR [domain.schema.SchemaBootstrap] [localhost-startStop-1] Schema auto-update failed
org.alfresco.repo.domain.schema.SchemaBootstrap$LockFailedException
	at org.alfresco.repo.domain.schema.SchemaBootstrap.checkSchemaPatchScripts(SchemaBootstrap.java:1000)
	at org.alfresco.repo.domain.schema.SchemaBootstrap.onBootstrap(SchemaBootstrap.java:1685)
	at org.springframework.extensions.surf.util.AbstractLifecycleBean.onApplicationEvent(AbstractLifecycleBean.java:56)
{code}

This is the second node- that is not doing the DB schema initialization log. 
The problem is that this second node could get past all the "updateSchema" without any problem - because calling checkSchemaPatchScripts and the subsequent countAppliedPatches (that would return 0 and not throw a NoSchemaException  while the first node in the cluster just managed to create the tables - but has not reached the code that writes any (activiti) patches as being applied in the alf_applied_patch table ). 
Then the second node did a check on line SchemaBootstrap.onBootstrap(SchemaBootstrap.java:1685 to see if it managed to install all the required patches - well this second node should not do those checks - as it is not the one that does the initialization of the DB. Furthermore this second node should not continue to execute any more code(fire the SchemaAvailableEvent event) until the first node has checked that all the patches were applied. 
In order to achieve this synchronization - I am proposing to move those checkSchemaPatchScripts from the onBootstrap to the end of updateSchema. This way the node that is doing the DB bootsrap also does the checks, and the other node is not allowed to continue the code until those checks have been done and they were successful (meaning that the first node doing the initialization would have deleted the alf_bootstrap_lock table ).

PR #119, opened 2018-05-09
SEARCH-830, fix error handling to display error message from solr

michaelsuzukisagi:


src/main/java/org/alfresco/repo/search/impl/solr/SolrSQLJSONResultSet.java
```diff
@@ -83,7 +83,14 @@ public SolrSQLJSONResultSet(JSONObject json, BasicSearchParameters searchParamet
                 {
                     throw new RuntimeException("Unable to execute the query, this API requires InsightEngine.");
                 }
-                throw new RuntimeException("Unable to execute the query, error caused by: " + error);
+                if(searchParameters.getExtraParameters().containsKey("format"))
+                {
+
```
tunaaksoy: I think you should either add a comment here or check the negative case and leave the else block.

src/main/java/org/alfresco/repo/search/impl/solr/SolrQueryClient.java
```diff
@@ -38,7 +38,7 @@
  */
 public interface SolrQueryClient extends BeanFactoryAware, InitializingBean
 {
-    public ResultSet executeQuery(final BasicSearchParameters searchParameters, String language);
+    public ResultSet executeQuery(final SearchParameters searchParameters, String language);
```
tunaaksoy: Will this cause backwards compatibility issues?

michaelsuzukisagi: No as this is a new interface and SearchParameter precedes BasicSearchParameters

src/main/java/org/alfresco/repo/search/impl/solr/SolrSQLJSONResultSet.java
michaelsuzukisagi: Removed redundant if statement

PR #120, opened 2018-05-09
SEARCH-824 (Remove unsupported search subsystems and configurations f…

tunaaksoy:
…rom Repository)

### DIFF
--- a/src/main/resources/alfresco/repository.properties
+++ b/src/main/resources/alfresco/repository.properties
@@ -90,7 +90,7 @@ system.workflow.maxPooledTasks=-1
 # Use '0' for unlimited.
 system.workflow.maxGroupReviewers=0
 
-index.subsystem.name=noindex
+index.subsystem.name=solr6
 
 # ######################################### #
 # Index Tracking Configuration              #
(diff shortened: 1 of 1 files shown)


src/main/resources/alfresco/repository.properties
```diff
@@ -834,12 +834,12 @@ encryption.reencryptor.numThreads=2
 
 # SOLR connection details (e.g. for JMX)
 solr.host=localhost
-solr.port=8080
-solr.port.ssl=8443
+solr.port=8983
+#solr.port.ssl=8983
```
michaelsuzukisagi: I would recommend using a different port in keeping with documentation.
It also looks weird in the ui to see 8983 for ssl and non ssl

tunaaksoy: I could not find any mention of a different port for SSL and after talking to you my understanding was that it might be the same port. That's the reason why it is set to 8983 but commented out.

michaelsuzukisagi: It just look strange when you look at the ui and see that both are on the same port.

tunaaksoy: Update the port number to 8984 (https://lucene.apache.org/solr/guide/6_6/enabling-ssl.html)

tunaaksoy:
Not so sure about the value for index.subsystem.name here:

https://github.com/Alfresco/alfresco-repository/blob/8b6d3ed16a0266039dc243d2020c7f717362b5e2/src/main/resources/alfresco/repository.properties#L93

Why was it set to "noindex"?

tunaaksoy:
Please also be aware that now SSL is not turn on by default ("solr.secureComms" is set to "none"). It has been decided that we won't change anything in the installer.

PR #124, opened 2018-05-23
REPO-3548: ACE-5852: Unable to create dynamic model as system user

andrei-rebegea:


eknizat:
LGTM, I had to look at the diff in intelij though as the diff on the PR looks like you have changed a lot of code, maybe you have different (fixed) line endings? 

One very minor thing, the method name isModelAdmin could be something more generic now that it also checks for system user

PR #130, opened 2018-06-14
ALF-21992 fix BehaviourFilter check subClass QName

nbarithel:
Fix and unit tests

Change-Id: I0edc09ec36f4918bcc14a9ed57ad4a099c9a5ae8


src/test/java/org/alfresco/repo/policy/PolicyComponentTransactionTest.java
```diff
@@ -431,6 +443,31 @@ public void behaviourHierarchyTestWork(QName createDocType, ClassFilter... disab
...
+            }
+            finally
+            {
+                enableBehaviours(nodeRef, disableTypes);
+            }
+            transaction.commit();
+        }
+        catch(Exception e)
+        {
+            try { transaction.rollback(); } catch (IllegalStateException ee) {}
```
CodrinChirica: This try catch block shouldn't be on a single line. I'll merge it and format the code in the following commit.

PR #131, opened 2018-06-14
ALF-22006 fix disabling version policies on a node

nbarithel:
fix and unit tests.

Change-Id: I5685952d62e272d6ea1c2052e7e4eb29c407b281

### DIFF
--- a/src/main/java/org/alfresco/repo/version/common/AbstractVersionServiceImpl.java
+++ b/src/main/java/org/alfresco/repo/version/common/AbstractVersionServiceImpl.java
@@ -141,10 +141,10 @@ public abstract class AbstractVersionServiceImpl
     {
         // invoke for node type
         QName nodeTypeQName = nodeService.getType(nodeRef);
-        this.afterVersionRevertDelegate.get(nodeTypeQName).afterVersionRevert(nodeRef, version);
+        this.afterVersionRevertDelegate.get(nodeRef, nodeTypeQName).afterVersionRevert(nodeRef, version);
         // invoke for node aspects
         Set<QName> nodeAspectQNames = nodeService.getAspects(nodeRef);
-        this.afterVersionRevertDelegate.get(nodeAspectQNames).afterVersionRevert(nodeRef, version);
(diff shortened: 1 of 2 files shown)


derekhulley:
Thank you for the contribution.  It has value and we will be happy to incorporate it.
In order to do so, I believe there are two considerations:
- this represents an API change, in effect.  Can the contribution be changed to extend the API so that existing code remain unchanged?
- The policy behaviour enable/disable is transaction- and thread-local.  Therefore, it is possible to achieve the same effect as long as you use try-finally when enabling or disabling behaviour.  Could you add details to the ALF issue on the problem you are trying to solve, please?

nbarithel:
Hi @derekhulley,

- I do not see why this would qualify as an API change. The fact that version related policies cannot be disabled on a specific node by calling the right API (see ALF-22006 policyBehaviourFilter.disableBehaviour(nodeRef, TYPE_QNAME)) but can be for all nodes is an unexpected bug...
- This is indeed the workaround but it is not thorough in case it is needed to disable the policy early in the execution before running third party code that would create versions on other documents.

Description updated in the ALF issue.

derekhulley:
You may regard the current API as deficient.  However, there will be extensions out there that make use of it the way it is.  Those extensions will break if we accept the changes as they are; it is an API change.

nbarithel:
I am not sure about which method you want to deprecate, the VersionServicePolicies ? They are working fine but just cannot be disabled on a particular node. I do not see how one would disable policies on a node (and on a specific className), let's say one of the NodeServicePolicies, but would require that VersionServicePolicies were still executed. It is an eventuality (probably not on purpose) but an unexpected bug...

Moreover I do not see how an API change would help in understanding the problem, could you elaborate ?

derekhulley:
Having a bit more of a look, I think you're right.  My concern was with the implicit contract for client code using the behaviours.  Actually your tests seem to cover that quite well.
We'll pick up the JIRA issue and hopefully get it looked at soon.

nbarithel:
Great ! I am looking forward to seeing the merge.

PR #132, opened 2018-06-15
REPO-3572: HotFix: MNT-19570: Multi-threaded check-in and check-out throwing exception using CMIS API

cristianturlica:


skopf:
Looks good to me

PR #134, opened 2018-06-15
REPO-3587: Add solr-search-context.xml startup test

eknizat:


src/test/java/org/alfresco/repo/search/SolrSearchContextTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
ancutam: not that important, but 2018 :)

PR #135, opened 2018-06-18
EGATE-55 move alfresco events repo

jamalkm:
EGATE-35/EGATE-55:
 - Moved with history 'alfresco-events-repo' and 'alfresco-messaging-repo' components to alfresco-repository project.
 - Move clean-up (removed duplicate folders)
 - Fixed compilation errors and updated the copyright headers
 - Cleaned-up pom.xml
 - Fixed minor code standard issues - convert tabs to spaces


derekhulley:
@jamalkm  What is the level of unit test coverage for the new packages?

jamalkm:
@derekhulley The alfresco-messaging-repo has only two Java classes for route configuration and the current tests cover that. On the other hand, even though alfresco-events-repo has only a couple of unit tests (~10% coverage), the proper tests (integration tests) are performed by underlying services like Sync service.

derekhulley:
Thanks for the update, @jamalkm.
We would need to add integration tests for the repository, then.   The events are a contract and we should therefore ensure they are correct, independently.  That's not in scope of this work - I've noted it to be sure we don't forget to address it at some point.

derekhulley:
This change is too big and the timing too critical to integrate onto master, just yet.
Please wait until the ACS 6.0 release is done.  The same is true of the packaging and sync projects, that need to be released for ACS 6.0 and be made ready for the next set of changes, such as this.

cristianturlica:
After reviewing the PR in the repo team the following changes are required:

derekhulley:
We can rename the properties at a later date once we understand the low-level events.

jamalkm:
Thanks for the review. I'll Merge it in 10-15mins.

PR #141, opened 2018-07-03
REPO-3259

skopf:
See https://issues.alfresco.com/jira/browse/REPO-3259?focusedCommentId=575518&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-575518
(undisclosed ticket; not visible outside of Alfresco)


skopf:
Indeed. That's not really great - if not even bad.
I had the same concerns and was looking for a way to prevent this. But there is no similar method in the AuthorityDAO that I could use to detect if the current user is an Admin. I would need to replicate code from the AuthorityService in the AuthorityDAO which would be even worse.
Another possibility would be to separate the Behaviour out in it's own class. But that's bad as well. This behaviour is closely connected to the actual data model, which is implemented by this DAO.

dcaruana:
The circular dependency is indeed undesired. Having looked at the code as well, it does seem difficult to remove this dependency with the addition of an admin authority check. Are we sure that the admin check is the solution?

skopf:
I am looking at an alternate solution for this problem and will soon start a new PR from a different branch. Then we can decide between both solutions.

PR #142, opened 2018-07-04
REPO-3539 : Integrate Api-Explorer

Epurashu:
Add api-explorer.url property in repository.property
   Add api-explorer.url property in sysadmin-parameter-context to be loaded at system startup (in SysAdminParams)
   Add apiExplorerUrl variable to SysAdminParamsImpl, setters and getters also add getApiExplorerUrl to SysAdminParams(interface)
   Add getApiExplorerUrl implementation in UrlUtil used for generating the REST Api-Explorer url used in /alfresco page(Index.jsp)


src/main/java/org/alfresco/util/UrlUtil.java
```diff
@@ -70,6 +70,34 @@ public static String getShareUrl(SysAdminParams sysAdminParams)
...
+     * @return Rest-Api Url such as https://col.ab.or.ate/api-explorer/
+     *  or http://localhost:8082/api-explorer/
+     */
+    public static String getApiExplorerUrl(SysAdminParams sysAdminParams, String requestURL, String requestURI)
+    {
+        if (!sysAdminParams.getApiExplorerUrl().isEmpty())
+        {
+            return sysAdminParams.getApiExplorerUrl();
+        }
+        else
```
andrei-rebegea: this else is not really required

src/main/java/org/alfresco/util/UrlUtil.java
```diff
@@ -70,6 +70,34 @@ public static String getShareUrl(SysAdminParams sysAdminParams)
...
+        {
+            if (!requestURI.isEmpty() && !requestURL.isEmpty())
+            {
+                return requestURL.replace(requestURI,"/api-explorer");
+            }
+        }
+        return new String();
```
andrei-rebegea: you can just use "". it may help the compiler with the immutability of the string objects and all

PR #143, opened 2018-07-05
(WIP) REPO-3594: Addition data points for acs.repository.info collector

ancutam:
- initial addition on db and system data points

### DIFF
--- a/src/main/java/org/alfresco/heartbeat/InfoDataCollector.java
+++ b/src/main/java/org/alfresco/heartbeat/InfoDataCollector.java
@@ -34,6 +34,10 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.InitializingBean;
 
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
 import java.util.*;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/alfresco/heartbeat/InfoDataCollector.java
```diff
@@ -126,7 +138,43 @@ public void afterPropertiesSet() throws Exception
...
+            DatabaseMetaData dbmeta = con.getMetaData();
+
+            infoValues.put("db.vendor", dbmeta.getDatabaseProductName());
+            infoValues.put("db.version", dbmeta.getDatabaseProductVersion());
+            infoValues.put("db.driverName", dbmeta.getDriverName());
+            infoValues.put("db.driverVersion", dbmeta.getDriverVersion());
+//            infoValues.put("db driverClass", dbmeta.getDriver); IS THERE ANY ? 
+        } 
+        catch (SQLException e)
+        { //TODO:
```
derekhulley: Do not leave a TODO here.  Is the exception worth logging or just a single-line WARN that the values could not be retrieved.

eknizat: In some of our other collectors, if we can't get a value for something, we just don't add it without logging anything. LicenseDataCollector is an example of this.

derekhulley: Then remove the TODO and put a comment to that effect, please.

PR #144, opened 2018-07-05
REPO-3617: Extend acs.repository.usage.system

eknizat:


src/main/java/org/alfresco/heartbeat/SystemUsageDataCollector.java
```diff
@@ -84,6 +111,40 @@ public void afterPropertiesSet() throws Exception
...
+        if (osMBean != null)
+        {
+            if (osMBean instanceof UnixOperatingSystemMXBean)
+            {
+                long openFileDescriptorCount = ((UnixOperatingSystemMXBean) osMBean).getOpenFileDescriptorCount();
+                systemUsageValues.put("openFileDescriptorCount", new Long(openFileDescriptorCount));
+            }
+
+            // processor info
+            double processCpuLoad = osMBean.getProcessCpuLoad() * 100;
```
derekhulley: Let's use an integer here.  A double can be very accurate and we'll end up sending long json values with little point.

PR #149, opened 2018-07-16
WIP: RM-6141: RM_support_for_the_Insight_Engine

tunaaksoy:


tunaaksoy:
SQL support (Insight Engine) for RM. Prototype implementation. MR will be close and the code will be kept as reference. Changes needed in other projects as well

MR: https://git.alfresco.com/search_discovery/InsightEngine/merge_requests/47
MR: https://git.alfresco.com/records-management/records-management/merge_requests/1085

src/main/java/org/alfresco/repo/search/impl/solr/SolrSQLHttpClient.java
```diff
@@ -200,6 +200,15 @@ public ResultSet executeQuery(SearchParameters searchParameters, String language
                 locales.put(I18NUtil.getLocale());
             }
             body.put("locales", locales);
+
+            // FIXME
```
michaelsuzukisagi: is this comment needed?

tunaaksoy: No, not really. I've put the same comment in all MRs I've created as it was not intended as production ready code (hence the MRs being closed). This MR is created only for future reference.

michaelsuzukisagi: Looks fine, do we have a TAS Rest API test to support this

meenalbhave: I think filter query / filter query parameter is not supported for /sql api yet, hence no tas rest api test is available for now.

michaelsuzukisagi: Yes, this is a note that we can merge once we have a test to support the change and a green build.

PR #150, opened 2018-07-17
REPO-3625 / MNT-19701: Groups rest api call is very expensive on the DB CPU

ancutam:
I changed the JOIN to LEFT JOIN to the `alf_store` table, in the statement that gets eventually executed when the groups api is called.
There have been other issues in the past related to the fact that the inner join slows down the query and the left join performs better in practice (some linked in the MNT).


montgolfiere:
Yes, although it's been there for many years. I wonder if snippet may cut-and-paste a couple of times so we should check for any other similar occurrences, if not already.

derekhulley:
We should be careful of trying to over-achieve.  If we get everything working well on latest, it will probably be fine.

montgolfiere:
Yes. I assume "working well on latest" means that this (and any additional) change does not cause measurable performance regression for other common DB use-cases.

derekhulley:
Are you suggesting, @montgolfiere, that we do the performance testing or dive deeper into the meaning of "working well", here?

PR #154, opened 2018-07-23
REPO-3720 MNT-19825 ImageMagick docker transformer does not take arbi…

alandavis:
…trary command line options.


src/main/java/org/alfresco/repo/content/transform/magick/ImageMagickContentTransformerWorker.java
```diff
@@ -374,7 +378,9 @@ protected void transformRemote(ContentReader reader, ContentWriter writer, Trans
                 "resizeHeight", resizeHeight,
                 "resizePercentage", resizePercentage,
                 "allowEnlargement", allowEnlargement,
-                "maintainAspectRatio", maintainAspectRatio);
+                "maintainAspectRatio", maintainAspectRatio,
+
```
killerboot: An extra line seems to be odd.

alandavis: Added a comment to say why.

PR #155, opened 2018-07-30
REPO-3626:  Add new option class TikaTransformationOptions. 

anechifor:
…We have set to true notExtractBookmarkTest for pdfBox.


src/main/java/org/alfresco/repo/content/transform/TikaPoweredContentTransformer.java
```diff
@@ -223,6 +225,14 @@ protected ParseContext buildParseContext(Metadata metadata, String targetMimeTyp
        }
        return context;
     }
+    
+    protected void buildExtraOptionsforTika(TransformationOptions options)
+    {
+        if (options == null)
+        {
+            options = new TransformationOptions();
```
alandavis: Should this be:  this.options =

alandavis: There is something wrong here. The newly created object is discarded. I think you need to return it from the method and then on the call do:

options = buildExtr...(options);

src/main/java/org/alfresco/repo/content/transform/PdfBoxContentTransformer.java
```diff
@@ -78,6 +79,14 @@ protected ParseContext buildParseContext(Metadata metadata, String targetMimeTyp
...
         return context;
     }
+    
+    @Override
+    protected void buildExtraOptionsforTika(TransformationOptions options)
+    {
+        TikaTransformationOptions tikaOptions = new TikaTransformationOptions();
+        tikaOptions.copyFrom(options);
+        tikaOptions.setNotExtractBookmarksText(true);
+    }
```
alandavis: See next comment about having to return the object.

src/main/java/org/alfresco/repo/content/transform/tika/TikaTransformationOptions.java
```diff
@@ -0,0 +1,63 @@
...
+     * <p/>
+     * The default is <code>false</code>
+     * 
+     * @param notExtractBookmarksText
+     */
+    public void setNotExtractBookmarksText(boolean notExtractBookmarksText)
+    {
+        this.notExtractBookmarksText = notExtractBookmarksText;
+    }
+
```
alandavis: I was expecting to see overridden toMap() and copyFrom() methods.

src/main/java/org/alfresco/repo/content/transform/TikaPoweredContentTransformer.java
```diff
@@ -290,9 +300,27 @@ protected void transformRemote(RemoteTransformerClient remoteTransformerClient,
     {
         String transform = getTransform();
         long timeoutMs = options.getTimeoutMs();
+        String notExtractBookmarksText = null;
+
+        buildExtraOptionsforTika(options);
```
alandavis: I might be wrong, but I don't think this approach is going to work. Don't there need to be toMap and copyFrom methods in the TransformationOptions object? By the time you get here the action service will have thrown away the bits of the map you are using. 

If it does work, I would remove the "forTika" from the name as it might get used in another transformer.

src/main/java/org/alfresco/repo/content/transform/TikaPoweredContentTransformer.java
```diff
@@ -290,9 +308,19 @@ protected void transformRemote(RemoteTransformerClient remoteTransformerClient,
     {
         String transform = getTransform();
         long timeoutMs = options.getTimeoutMs();
+        String notExtractBookmarksText = null;
+
+        if (extractBookmarksText)
+        {
```
alandavis: Is there a ! missing here in the if?

src/main/java/org/alfresco/repo/content/transform/TikaPoweredContentTransformer.java
```diff
@@ -223,6 +231,16 @@ protected ParseContext buildParseContext(Metadata metadata, String targetMimeTyp
...
+    
+    protected TransformationOptions buildExtraOptions(TransformationOptions options)
+    {
+        if (options == null)
+        {
+            options = new TransformationOptions();
+        }
+        
+        return options;
+    }
```
alandavis: Don't think this method is needed any more.

src/main/java/org/alfresco/repo/content/transform/TikaPoweredContentTransformer.java
```diff
@@ -90,6 +91,8 @@
     protected static final String LINE_BREAK = "\r\n";
     public static final String WRONG_FORMAT_MESSAGE_ID = "transform.err.format_or_password";
     
+    private boolean extractBookmarksText = true;
```
alandavis: This instance variable (and all the code that uses it) should be in the sub class PdfBoxContentTransformer.
This will mean that you need to override the transformRemote method just like ArchiveContentTransformer does.

alandavis:
LGTM

PR #161, opened 2018-08-21
REPO-3747 Trials: Changes to license generation and docs

alandavis:
Added log message: "To extend the trial period visit: https://www.alfresco.com/platform/content-services-ecm/trial/docker"

### DIFF
--- a/src/main/java/org/alfresco/repo/descriptor/DescriptorStartupLog.java
+++ b/src/main/java/org/alfresco/repo/descriptor/DescriptorStartupLog.java
@@ -132,10 +132,11 @@ public class DescriptorStartupLog extends AbstractLifecycleBean
             }
             
             Date validUntil = license.getValidUntil();
-            
+
+            Integer days = null;
             if (validUntil != null)
             {
-                Integer days = license.getDays();
+                days = license.getDays();
                 Integer remainingDays = license.getRemainingDays();
                 
                 msg += " limited to " + days + " days expiring " + validUntil + " (" + remainingDays + " days remaining).";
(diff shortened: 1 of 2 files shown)


src/main/resources/alfresco/messages/system-messages.properties
```diff
@@ -114,3 +114,4 @@ system.license.err.reloadFailed=Failed to reload license: {0}
 # Startup message
 system.info.startup=Alfresco Content Services started ({0}{1}{2}). Current version: {3} schema {4}. Originally installed version: {5} schema {6}.
 system.warn.readonly=Alfresco Content Services is currently in read-only mode. Check your license.
+system.info.limited_trial=Note: this is a limited trial of the Enterprise version of Alfresco Content Services that goes into read-only mode after 2-days. Request an extended 30-day trial at: https://www.alfresco.com/platform/content-services-ecm/trial/docker
```
eknizat: Do you think it is worth taking the url out to somewhere else? If so, where would be a good place for it?

src/main/java/org/alfresco/repo/descriptor/DescriptorStartupLog.java
```diff
@@ -1,28 +1,28 @@
...
- * 
- * You should have received a copy of the GNU Lesser General Public License
- * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
- * #L%
- */
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
CodrinChirica: Minor observation: should the version be bumped to 2018?

PR #166, opened 2018-09-03
REPO-3713: Raw events: onContentUpdate

cristianturlica:
- added Event behavior and tests
   - test that relies on activeMQ to be up is disabled for now
   - enabled messaging.subsystem

### DIFF
--- a/src/main/java/org/alfresco/repo/policy/EventBehaviour.java
+++ b/src/main/java/org/alfresco/repo/policy/EventBehaviour.java
@@ -35,32 +35,38 @@ import org.alfresco.repo.rawevents.EventProducer;
 import org.springframework.extensions.surf.util.ParameterCheck;
 
 /**
- * Event based Behaviour.
+ * Event based Behaviour. <br/>
+ * <p/>
+ * A client uses an <code>EventBehaviour</code> to bind a send event behaviour to a Class-level Policy.
+ * <br/>
+ * <p/>
+ * The event behavior delegates the generation of the event to a method pointer.
+ * The pointer is represented by an instance object and method name.
+ * <p/>
  * 
- * The event behavior delegates the generation of the event to a method pointer.  The pointer is
- * represented by an instance object and method name.
- *
  * @author Cristian Turlica
  */
 @AlfrescoPublicApi
 public class EventBehaviour extends BaseBehaviour
 {
 
-    EventProducer eventProducer;
+    private EventProducer eventProducer;
 
-    String endpointUri;
+    private String endpointUri;
 
     // The object instance holding the method
-    Object instance;
+    private Object instance;
 
     // The method name
-    String method;
+    private String method;
 
     /**
      * Construct.
      * 
-     * @param instance  the object instance holding the method
-     * @param method  the method name
+     * @param instance
+     *            the object instance holding the method
+     * @param method
+     *            the method name
      */
     public EventBehaviour(EventProducer eventProducer, String endpointUri, Object instance, String method)
     {
@@ -90,35 +96,38 @@ public class EventBehaviour extends BaseBehaviour
         this.method = method;
     }
 
-
     @Override
     public String toString()
     {
         return "Java method[class=" + instance.getClass().getName() + ", method=" + method + "]";
     }
-    
+
     @SuppressWarnings("unchecked")
-	public synchronized <T> T getInterface(Class<T> policy) 
-	{
-	    ParameterCheck.mandatory("Policy class", policy);
-	    Object proxy = proxies.get(policy);
-	    if (proxy == null)
-	    {
-	        InvocationHandler handler = getInvocationHandler(eventProducer, endpointUri, instance, method, policy);
-	        proxy = Proxy.newProxyInstance(policy.getClassLoader(), new Class[]{policy}, handler);
-	        proxies.put(policy, proxy);
-	    }
-	    return (T)proxy;
-	}
+    public synchronized <T> T getInterface(Class<T> policy)
+    {
+        ParameterCheck.mandatory("Policy class", policy);
+        Object proxy = proxies.get(policy);
+        if (proxy == null)
+        {
+            InvocationHandler handler = getInvocationHandler(eventProducer, endpointUri, instance, method, policy);
+            proxy = Proxy.newProxyInstance(policy.getClassLoader(), new Class[] { policy }, handler);
+            proxies.put(policy, proxy);
+        }
+        return (T) proxy;
+    }
 
     /**
      * Gets the Invocation Handler.
      * 
-     * @param <T>  the policy interface class
-     * @param instance  the object instance
-     * @param method  the method name
-     * @param policyIF  the policy interface class  
-     * @return  the invocation handler
+     * @param <T>
+     *            the policy interface class
+     * @param instance
+     *            the object instance
+     * @param method
+     *            the method name
+     * @param policyIF
+     *            the policy interface class
+     * @return the invocation handler
      */
     <T> InvocationHandler getInvocationHandler(EventProducer eventProducer, String endpointUri, Object instance, String method, Class<T> policyIF)
     {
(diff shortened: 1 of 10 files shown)


src/main/java/org/alfresco/repo/policy/EventBehaviour.java
```diff
@@ -0,0 +1,242 @@
...
+
+
+    @Override
+    public String toString()
+    {
+        return "Java method[class=" + instance.getClass().getName() + ", method=" + method + "]";
+    }
+    
+    @SuppressWarnings("unchecked")
+	public synchronized <T> T getInterface(Class<T> policy)
```
derekhulley: Format

src/main/java/org/alfresco/repo/policy/EventBehaviour.java
```diff
@@ -0,0 +1,242 @@
...
+            Method delegateMethod = instanceClass.getMethod(method, (Class[])policyIFMethods[0].getParameterTypes());
+            return new JavaMethodInvocationHandler(eventProducer, endpointUri, this, delegateMethod);
+        }
+        catch (NoSuchMethodException e)
+        {
+            throw new PolicyException("Method " + method + " not found or accessible on " + instance.getClass(), e);
+        }
+    }    
+    
+    /**
```
derekhulley: Please elaborate on the class's function.

src/main/java/org/alfresco/repo/policy/EventBehaviour.java
```diff
@@ -0,0 +1,242 @@
...
+         */
+        private JavaMethodInvocationHandler(EventProducer eventProducer, String endpointUri, EventBehaviour behaviour, Method delegateMethod)
+        {
+            this.eventProducer = eventProducer;
+            this.endpointUri = endpointUri;
+            this.behaviour = behaviour;
+            this.delegateMethod = delegateMethod;
+        }
+
+        /* (non-Javadoc)
```
derekhulley: (non-Javadoc) is just a waste of screen space and does not provide any information that is not available in an IDE.  Please turn off this auto-generation in IDE.

src/main/java/org/alfresco/repo/rawevents/types/Event.java
```diff
@@ -0,0 +1,44 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.repo.rawevents.types;
+
+/**
+ * This interface defines the basic data which are captured for each event.
```
derekhulley: "which is captured"

src/main/java/org/alfresco/repo/rawevents/types/EventType.java
```diff
@@ -0,0 +1,33 @@
...
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.repo.rawevents.types;
+
+import java.io.Serializable;
+
+public enum EventType implements Serializable
+{
+    CONTENT_UPDATED
```
derekhulley: Comments on the enum and each entry, please.

src/main/java/org/alfresco/repo/rawevents/types/OnContentUpdatePolicyEvent.java
```diff
@@ -0,0 +1,52 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.repo.rawevents.types;
+
+public class OnContentUpdatePolicyEvent extends EventImpl
```
derekhulley: Description, please.

src/main/java/org/alfresco/repo/policy/EventBehaviour.java
```diff
@@ -0,0 +1,242 @@
...
+ * The event behavior delegates the generation of the event to a method pointer.  The pointer is
+ * represented by an instance object and method name.
+ *
+ * @author Cristian Turlica
+ */
+@AlfrescoPublicApi
+public class EventBehaviour extends BaseBehaviour
+{
+
+    EventProducer eventProducer;
```
killerboot: Are these variables intentionally package-private?

src/main/java/org/alfresco/repo/rawevents/EventProducer.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+
+import org.alfresco.error.AlfrescoRuntimeException;
+import org.apache.activemq.transport.amqp.message.AmqpMessageSupport;
+import org.apache.camel.ProducerTemplate;
+import org.apache.commons.lang.StringUtils;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class EventProducer
```
killerboot: This class needs JavaDoc description.

src/main/resources/alfresco/messaging-context.xml
```diff
@@ -39,4 +39,13 @@
             </list>
         </property>
     </bean>
+
+    <bean id="messagingObjectMapper" class="org.gytheio.messaging.jackson.ObjectMapperFactory"
```
killerboot: Can we swap that to [Jackson2ObjectMapperFactoryBean](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/converter/json/Jackson2ObjectMapperFactoryBean.html) from Spring?

Are we ok with reusing the Messaging subsystem from Desktop Sync?
I presume Camel configs are shared as well.

src/test/java/org/alfresco/repo/rawevents/EventBehaviorTest.java
```diff
@@ -0,0 +1,421 @@
...
+                this.policyComponent.removeClassDefinition(classBehaviour);
+            }
+        }
+
+    }
+
+    /**
+     * Tests that the content update policy is triggered correctly for every event.
+     */
+    // @Test
```
killerboot: We can spin up ActiveMQ in Docker with PostreSQL for tests.
I can see it as a separate Job in Bamboo.

src/test/java/org/alfresco/repo/rawevents/EventBehaviorTest.java
```diff
@@ -0,0 +1,421 @@
...
+        {
+            authenticationComponent.clearCurrentSecurityContext();
+        }
+        catch (Throwable e)
+        {
+            // ignore
+        }
+    }
+
+    private UserTransaction getUserTransaction()
```
killerboot: Please use `@Transactional` instead for cleaner code. It rollbacks by default.
If required RetryingTransactionHelper is a better choice as some of the tests might intermittently fail without retries.

skopf:
Some more comments:
1. There are policies that define a return value. We need to handle this, e.g. by throwing an Exception when someone tries to register such an {{EventBehaviour}} for this policy
2. At least we should avoid using Reflection here. This is always a source of runtime errors as the compiler is not able to check method signatures
3. Instead, we could define an interface like {{MessageGenerator}} that defines one method like {{Object generateMessage(String policy, Object[] args)}}

src/main/java/org/alfresco/repo/policy/EventBehaviour.java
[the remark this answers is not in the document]
cristianturlica: Changed thanks.

src/main/java/org/alfresco/repo/policy/EventBehaviour.java
cristianturlica: Changed to private for now.

src/main/java/org/alfresco/repo/rawevents/EventProducer.java
cristianturlica: Added, thanks.

src/main/resources/alfresco/messaging-context.xml
cristianturlica: I added Jackson2ObjectMapperFactoryBean, thanks for the recommendation, we removed the use of gytio now.

src/test/java/org/alfresco/repo/rawevents/EventBehaviorTest.java
cristianturlica: Re-enabled the test. We will need to update the bamboo job.

src/test/java/org/alfresco/repo/rawevents/EventBehaviorTest.java
cristianturlica: We can't use @Transaction for now because we need OwnJVMTestsCategory... I used RetryingTransactionHelper instead.

cristianturlica:
Changes done.

skopf:
You didn't push anything to the branch

cristianturlica:
Sorry, please see now.

skopf:
Looks good now, really great work.

PR #168, opened 2018-09-10
REPO-3688: Deprecate public APIs

eknizat:
* Deprecate all classes in the following packages and their subpackages:
org.alfresco.repo.content.transform
org.alfresco.repo.rendition (not deprecated RenditionPreventionRegistry)
org.alfresco.service.cmr.rendition
org.alfresco.repo.thumbnail
org.alfresco.service.cmr.thumbnail

* Deprecate transformation related methods in ContentService and ContentServiceImpl

* Deprecated other classes used for local transformations or renditions:
AbstractTransformationSourceOptions
SerializedTransformationOptionsAccessor
TemporalSourceOptions
TransformationSourceOptions
CMISRenditionMapping

* Deprecate all other methods where the classes from the deprecated packages are used:
AbstractContentReader - Deprecated methods relating to transformations
ScriptNode - Deprecated thumbnail, transformation and Rendering API methods
BaseContentNode - Deprecated getContentAsText() as it uses deprecated transformation methods
CMISConnector - Deprecated methods relating to transformations

* Deprecated methods in ServiceRegistry and ServiceDescriptorRegistry that return deprecated services

* Deprecate related junit tests

### DIFF
--- a/src/main/java/org/alfresco/opencmis/CMISConnector.java
+++ b/src/main/java/org/alfresco/opencmis/CMISConnector.java
@@ -309,6 +309,7 @@ public class CMISConnector implements ApplicationContextAware, ApplicationListen
     private CheckOutCheckInService checkOutCheckInService;
     private LockService lockService;
     private ContentService contentService;
+    @Deprecated
     private RenditionService renditionService;
     private FileFolderService fileFolderService;
     private TenantAdminService tenantAdminService;
@@ -327,6 +328,7 @@ public class CMISConnector implements ApplicationContextAware, ApplicationListen
     private DictionaryService dictionaryService;
     private SiteService siteService;
     private ActionService actionService;
+    @Deprecated
     private ThumbnailService thumbnailService;
     private ServiceRegistry serviceRegistry;
     private EventPublisher eventPublisher;
@@ -498,6 +500,12 @@ public class CMISConnector implements ApplicationContextAware, ApplicationListen
         return openHttpSession;
     }
 
+    /**
+     *
+     *
+     * @deprecated The thumbnails code is being moved out of the codebase and replaced by the new async RenditionService2 or other external libraries.
+     */
+    @Deprecated
     public void setThumbnailService(ThumbnailService thumbnailService)
     {
 		this.thumbnailService = thumbnailService;
@@ -614,7 +622,10 @@ public class CMISConnector implements ApplicationContextAware, ApplicationListen
 
     /**
      * Sets the rendition service.
+     *
+     * @deprecated The RenditionService is being replace by the simpler async RenditionService2.
      */
+    @Deprecated
     public void setrenditionService(RenditionService renditionService)
     {
         this.renditionService = renditionService;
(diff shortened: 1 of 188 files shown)


src/main/java/org/alfresco/repo/content/AbstractContentReader.java
```diff
@@ -374,7 +398,10 @@ public void contentStreamClosed() throws ContentIOException
 
     /**
      * @see Channels#newInputStream(java.nio.channels.ReadableByteChannel)
+     *
+     * @deprecated The transformations code is being moved out of the codebase and replaced by the new async RenditionService2 or other external libraries.
      */
+    @Deprecated
```
derekhulley: We are *not* going to remove this method, so the deprecation here is not correct.
Where an implementation uses deprecated methods, it should not automatically be marked as deprecated.
You do not have to mark as ignored (internally) but you can.

src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -970,7 +981,10 @@ protected void onShutdown(ApplicationEvent event)
      * Asynchronously generates thumbnails for the given node.
      *
      * @param nodeRef NodeRef
+     *
+     * @deprecated The thumbnails code is being moved out of the codebase and replaced by the new async RenditionService2 or other external libraries.
      */
+    @Deprecated
     public void createThumbnails(NodeRef nodeRef, Set<String> thumbnailNames)
```
derekhulley: This method is a function that does not required any specific implementation.  We will be re-implementing, as required.

derekhulley: This one is tricky; you'd have to look where it's used as to whether we deprecate speculatively.

src/main/java/org/alfresco/repo/jscript/ScriptNode.java
```diff
@@ -2706,7 +2706,10 @@ public ScriptNode cancelCheckout()
      * @param mimetype   Mimetype destination for the transformation
      * 
      * @return Node representing the newly transformed document.
+     *
+     * @deprecated The transformations code is being moved out of the codebase and replaced by the new async RenditionService2 or other external libraries.
      */
+    @Deprecated
     public ScriptNode transformDocument(String mimetype)
```
derekhulley: This functionality will be preserved and the implementation will change.
This ScriptNode is a public API and we have to preserve the features it provides.

src/main/java/org/alfresco/repo/template/BaseContentNode.java
```diff
@@ -588,7 +588,11 @@ public String getContentMaxLength(int length)
          * 
          * @return the binary content stream converted to text using any available transformer
          *         if fails to convert then null will be returned
+         *
+         * @deprecated The transformations code is being moved out of the codebase and replaced by the new async
+         * RenditionService2 or other external libraries.
          */
+        @Deprecated
         public String getContentAsText(int length)
```
derekhulley: Same, we'd just change the implementation.

eknizat:
I’m still not sure about CMISRenditionMapping and CMISConnecotr.getRenditionMapping(), do you think we should keep those?

alandavis:
Wow what lot of deprecation. :)

You probable also need to deprecate the support tool's "Test Transform" code in the Admin console.

The Admin console also has a lot of transformer settings that will need to be removed at some point, so should be deprecated too.

PR #169, opened 2018-09-20
REPO-3680: Renditions: Rendition Service can call new Transform Service

cristianturlica:
- added send after commit support


src/main/java/org/alfresco/repo/rawevents/EventProducer.java
```diff
@@ -51,8 +59,12 @@
  */
 public class EventProducer
 {
+    private static final String POST_TRANSACTION_PENDING_REQUESTS = "postTransactionPendingEventRequests";
     protected static final String ERROR_SENDING = "Could not send event";
 
+    private static Log logger = LogFactory.getLog(EventProducer.class);
+
+    private TransactionListener transactionListener = new TransactionListener();
```
killerboot: It is the same bug (and lot's of code duplication), that was introduced in this [branch](https://github.com/Alfresco/alfresco-repository/tree/feature/repo-3676-another-approach) in RendtionService2Impl.

The same tx listener will be bound to all tx. That may result in concurrency issues.
I corrected it in a refactoring, see 68d66b1905dbbcaf38bbba5957cd1eb39139eeae. This needs to go through testing. But the new RenditionRequestScheduler can be generic enough to execute any post transaction code.

src/test/java/org/alfresco/repo/rawevents/EventProducerTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Provides a base set of tests for {@link EventProducer}
+ *
+ * @author Cristian Turlica
+ */
+@Category(OwnJVMTestsCategory.class)
+public class EventProducerTest
```
killerboot: Is this included in any test suits?

Epurashu: It appears to not be part of any test suite, please add it 👍 .

cristianturlica: Changed, thanks.

PR #174, opened 2018-09-27
Feature/repo 3766

alandavis:
Use RenditionService2 when Share creates an initial thumbnail.

### DIFF
--- a/src/main/java/org/alfresco/repo/jscript/ScriptNode.java
+++ b/src/main/java/org/alfresco/repo/jscript/ScriptNode.java
@@ -151,7 +151,7 @@ import org.springframework.extensions.surf.util.URLEncoder;
 public class ScriptNode implements Scopeable, NamespacePrefixResolverProvider
 {
     private static final long serialVersionUID = -3378946227712939601L;
-    public static final QName RENDITION_SERVICE_2 = QName.createQName(NamespaceService.ALFRESCO_URI, "RenditionService2");
+    private static final QName RENDITION_SERVICE_2 = QName.createQName(NamespaceService.ALFRESCO_URI, "RenditionService2");
 
     private static Log logger = LogFactory.getLog(ScriptNode.class);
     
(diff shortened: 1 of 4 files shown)


src/main/java/org/alfresco/repo/jscript/ScriptNode.java
```diff
@@ -272,6 +278,8 @@ public ScriptNode(NodeRef nodeRef, ServiceRegistry services, Scriptable scope)
         this.fileFolderService = services.getFileFolderService();
         this.retryingTransactionHelper = services.getTransactionService().getRetryingTransactionHelper();
         this.scope = scope;
+        renditionService2 = (RenditionService2)services.getService(RENDITION_SERVICE_2);
```
killerboot: Is it possible to add RenditionsService2 to the `ServiceRegistry` and call `services.getRenditionService2()` ?

alandavis: Just have, but looking up services this way no longer seen as the correct approach.

src/main/java/org/alfresco/repo/jscript/ScriptNode.java
```diff
@@ -3113,10 +3120,18 @@ public ScriptThumbnail createThumbnail(String thumbnailName, boolean async, bool
             }
             else
             {
-                Action action = ThumbnailHelper.createCreateThumbnailAction(details, services);
-                
-                // Queue async creation of thumbnail
-                this.services.getActionService().executeAction(action, this.nodeRef, true, true);
+                RenditionDefinition2 renditionDefinition = renditionDefinitionRegistry2.getRenditionDefinition(thumbnailName);
+                if (renditionDefinition != null)
```
killerboot: Is it worth checking if renditionService2 is null as well? In case if it was not injected it a custom code.

alandavis: I think that is going too far.

derekhulley:
Please can we have better MR titles.  "Feature/repo 3766" is not descriptive and does not play nicely with JIRA.

PR #175, opened 2018-09-27
REPO-3692: Add enterprise routs package to scanned packages

eknizat:
Add enterprise routs package to scanned packages for camel messaging

### DIFF
--- a/src/main/resources/alfresco/subsystems/Messaging/default/messaging-context.xml
+++ b/src/main/resources/alfresco/subsystems/Messaging/default/messaging-context.xml
@@ -68,6 +68,9 @@
     <!-- Any RouteBuilder in this package will be fired up -->
     <context:component-scan base-package="org.alfresco.messaging.camel.routes"/>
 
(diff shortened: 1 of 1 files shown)


src/main/resources/alfresco/subsystems/Messaging/default/messaging-context.xml
```diff
@@ -68,6 +68,9 @@
     <!-- Any RouteBuilder in this package will be fired up -->
     <context:component-scan base-package="org.alfresco.messaging.camel.routes"/>
 
+    <!-- Any RouteBuilder in this enterprise package will be fired up -->
+    <context:component-scan base-package="org.alfresco.enterprise.repo.rendition2.routes"/>
```
killerboot: As pointed by @cristianturlica the name might be more generic like `org.alfresco.enterprise.repo.routes`.

cristianturlica: We should change to "org.alfresco.enterprise.repo.routes" so we can reuse it also for other parts and not just renditions.

PR #177, opened 2018-10-04
REPO-3692: Fix thumbnail last modified logic

eknizat:
* The rendition service 2 now updates the lastThumbnailModification on the source node when the thumbnail is updated


alandavis:
LGTM

PR #182, opened 2018-10-09
MNT-19981 increase range of bootstrap lock

mmuller88:


andrei-rebegea:
this fix does not solve the problem of the activiti table initialization. The marker table is not checked in the if (!activitiTablesExist), therefore, even if you set the marker earlier... the code will still fail because the missing part is the check for the marker before creating the activiti tables

PR #183, opened 2018-10-09
REPO-3808: Fix activiti table bootstrap in large cluster and add more logging

andrei-rebegea:
- There was a check for the bootstrap marker missing in the if statement that initialized and created the activiti tables in the DB.
- Added more logging as the cluster boot up is much more frequent now with Kubernetes and docker, we need better logging so we can determine if there are problems, what went wrong

### DIFF
--- a/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
+++ b/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
@@ -918,7 +918,7 @@ public class SchemaBootstrap extends AbstractLifecycleBean
             if (logger.isInfoEnabled())
             {
                 logger.info(
-                    "Update case: checkSchemaPatchScripts scripts executed in " + (System.currentTimeMillis() - start) + " ms");
+                    "Update scripts executed in " + (System.currentTimeMillis() - start) + " ms");
             }
         }
 
@@ -992,7 +992,7 @@ public class SchemaBootstrap extends AbstractLifecycleBean
 
             if (logger.isInfoEnabled())
             {
-                logger.info("Update case: checking all patches are applied, executed in: " + (System.currentTimeMillis() - start)
+                logger.info("Checking all patches have been applied, executed in: " + (System.currentTimeMillis() - start) 
                     + " ms");
             }
         }
@@ -1727,11 +1727,16 @@ public class SchemaBootstrap extends AbstractLifecycleBean
                 {
                     try
                     {
+                        long start = System.currentTimeMillis();
+
                         createdSchema = updateSchema(connection);
                         updatedSchema = true;
+
                         if (logger.isInfoEnabled())
                         {
-                            logger.info("Created schema value: " + createdSchema);
+                            logger.info(
+                                (createdSchema ? "Created" : "Updated") +
+                                    " schema in " + (System.currentTimeMillis() - start) + " ms");
                         }
                         break;
                     }
--- a/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
+++ b/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
@@ -928,7 +928,7 @@ public class SchemaBootstrap extends AbstractLifecycleBean
         boolean activitiTablesExist = checkActivitiTablesExist(connection);
         if (logger.isInfoEnabled())
         {
-            logger.info("Activiti tables exist: " + activitiTablesExist);
+            logger.info("Activiti tables need to be " + (activitiTablesExist ? "checked for patches" : " created"));
         }
 
         if(!activitiTablesExist)
@@ -992,8 +992,8 @@ public class SchemaBootstrap extends AbstractLifecycleBean
 
             if (logger.isInfoEnabled())
             {
-                logger.info("Checking all patches have been applied, executed in: " + (System.currentTimeMillis() - start) 
-                    + " ms");
+                logger.info("Checking all patches have been applied, executed in: " +
+                             (System.currentTimeMillis() - start) + " ms");
             }
         }
 
--- a/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
+++ b/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
@@ -1114,7 +1114,7 @@ public class SchemaBootstrap extends AbstractLifecycleBean
     {
         if (isAnotherClusterMemberBootstrapping(connection))
         {
-            logger.info("Another Alfresco cluster node is updating the DB. Waiting for the other node to finish...");
+            logger.info("Another Alfresco cluster node is updating the DB");
 
             // We throw a well-known exception to be handled by retrying code if required
             throw new LockFailedException();
@@ -1731,14 +1731,16 @@ public class SchemaBootstrap extends AbstractLifecycleBean
                         if (logger.isInfoEnabled())
                         {
                             logger.info(
-                                (createdSchema ? "Created" : "Updated") +
-                                    " schema in " + (System.currentTimeMillis() - start) + " ms");
+                                (createdSchema ? "Creating" : "Updating") +
+                                    " the DB schema took " + (System.currentTimeMillis() - start) + " ms");
                         }
                         break;
                     }
                     catch (LockFailedException e)
                     {
-                        logger.info("DB lock failed on schema bootstrap. Attempt: " + i);
+                        logger.info(
+                            "The current Alfresco cluster node is waiting for another chance to bootstrap the DB schema. "
+                                + "Attempt: " + (i + 1) + " of " + schemaUpdateLockRetryCount);
                         try { this.wait(schemaUpdateLockRetryWaitSeconds * 1000L); } catch (InterruptedException ee) {}
                     }
                 }
--- a/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
+++ b/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
@@ -781,6 +781,10 @@ public class SchemaBootstrap extends AbstractLifecycleBean
         {
             stmt.executeUpdate("create table alf_bootstrap_lock (charval CHAR(1) NOT NULL)");
             // Success
+            if (logger.isInfoEnabled())
+            {
+                logger.info("Bootstrap started.");
+            }
             return;
         }
         catch (Throwable e)
@@ -799,9 +803,7 @@ public class SchemaBootstrap extends AbstractLifecycleBean
      * has created the alf_bootstrap_lock table - This table indicates that a schema initialization
      * or schema DB update is in progress.
      *
-     * Advice: Do not consider this locking system as bullet proof!
-     *
-     * @return true if the alf_bootstrap_lock exists; false otherwise
+     * @return true if the alf_bootstrap_lock marker table exists in the DB; false otherwise
      *
      */
     private synchronized boolean isBootstrapInProgress(Connection connection) throws Exception
@@ -811,6 +813,10 @@ public class SchemaBootstrap extends AbstractLifecycleBean
         {
             stmt.executeQuery("select * from alf_bootstrap_lock");
             // Success
+            if (logger.isInfoEnabled())
+            {
+                logger.info("Bootstrap marker still present in the DB.");
+            }
             return true;
         }
         catch (Throwable e)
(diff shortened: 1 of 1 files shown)


src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -811,6 +813,10 @@ private synchronized boolean isBootstrapInProgress(Connection connection) throws
         {
             stmt.executeQuery("select * from alf_bootstrap_lock");
             // Success
+            if (logger.isInfoEnabled())
```
killerboot: Does it need to be `info`?

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -884,22 +902,41 @@ private boolean updateSchema(Connection connection) throws Exception
...
             // Execute any pre-auto-update scripts
             checkSchemaPatchScripts(connection, preUpdateScriptPatches, true);
             
             // Execute any post-auto-update scripts
             checkSchemaPatchScripts(connection, postUpdateScriptPatches, true);
+
+            if (logger.isInfoEnabled())
+            {
+                logger.info(
+                    "Update case: checkSchemaPatchScripts scripts executed in " + (System.currentTimeMillis() - start) + " ms");
```
killerboot: Can it be rephrased like `Update scripts executed in ...`

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -884,22 +902,41 @@ private boolean updateSchema(Connection connection) throws Exception
...
+            }
         }
 
+        ensureCurrentClusterMemberIsBootstrapping(connection);
+
         // Initialise Activiti DB, using an unclosable connection
         boolean activitiTablesExist = checkActivitiTablesExist(connection);
+        if (logger.isInfoEnabled())
+        {
+            logger.info("Activiti tables exist: " + activitiTablesExist);
```
killerboot: The Activiti tables should exist on each of installation. Maybe change this one to `debug`?

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -884,22 +902,41 @@ private boolean updateSchema(Connection connection) throws Exception
...
+        }
+
         if(!activitiTablesExist)
         {
+            long start = System.currentTimeMillis();
+
+            ensureCurrentClusterMemberIsBootstrapping(connection);
```
killerboot: Why can't we check this only once for the whole bootstrap?

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -1645,6 +1729,10 @@ public synchronized void onBootstrap(ApplicationEvent event)
                     {
                         createdSchema = updateSchema(connection);
                         updatedSchema = true;
+                        if (logger.isInfoEnabled())
+                        {
+                            logger.info("Created schema value: " + createdSchema);
```
killerboot: This either needs to be a debug or say something that it is a fresh installation vs upgrade process.

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
andrei-rebegea: We could, but that would mean important refactoring of the SchemaBootstrap class. The code was design (as see in this comment from 2008:"// Retries are required here as the DB lock will be applied lazily upon first statement execution.") to back off only if changes are required to the db AND can't create the db lock maker(alf_bootstrap_lock).

How can you know in a node if changes are required?
Only by going through the initiation steps:
1. check if the ALF tables need to be created- and create them
2. check if the ALF tables need to be patched - and apply them
3. check if the Activiti tables are present - and create them
4. check if the Activiti tables need patches -and apply them

At what point can you determine if another node in the cluster is present? And tries to do the same thing? 
There is no hazelcast yet or any of the fancy DB lock mechanisms;

The way the code (with this additional fix) works is:  
1. all the nodes in the cluster can't wait around and assume that there is another node in the cluster that is doing the create/upgrade schema, so they, themselves, try to see if tables need to be created or updated
2. the *first node in the cluster that decides* it needs to do some changes in the DB, creates (if it can) that marker table (alf_bootstrap_lock) and marks in the object (executedStatementsThreadLocal) that it plans to execute the initialization/upgrade of the DB
3. all other nodes in the cluster, before executing ANY statement have to check if there is a marker in the DB telling them that some other node is doing the upgrade/initialization
4. Only one node should succeed in getting that alf_bootstrap_lock created, therefore only one node will do the initialization/upgrade
5. All other nodes will back off ONLY when they detect that changes are required; If changes are not required they will boot as fast/efficient as possible- not even caring if there are other nodes in the cluster at this point

Case "3. check if the Activiti tables are present" from above was not covered by these checks before this fix;
The problem with the Activiti table initialization (initialiseActivitiDBSchema) is that the code that actually creates the tables is in the Activiti code, and also the initialization of the objects required to create the Activiti tables takes a long time, and this helps with creating problems in the cluster synchronization(checks like checkActivitiTablesExist(connection); ) we have in the code;

Lets take the Activiti tables initialization ("3. check if the Activiti tables are present") example (2 nodes cluster for simplicity) BEFORE THIS FIX:
1. Both nodes boot up around the same time
2. First node was a little bit faster and finishes the ALF tables initialization (reaches line " boolean activitiTablesExist = checkActivitiTablesExist(connection);"); This means it was able to create the alf_bootstrap_lock marker table and also has some data in the executedStatementsThreadLocal variable;
3. Second node checks if the ALF tables are present- they are! and if they need upgrading - they don't! ; So the second node has not executed any statements in the DB up to now(executedStatementsThreadLocal should be empty);
4. Then both nodes check if the Activiti tables exist - they don't
5. The first node is starting initializing the Activiti beans and at some point creates the tables in the DB
6. (BEFORE THE FIX) the second node was doing the same thing, because there was no check or code to try to mark that this second node was trying to initialize the Activiti tables. It did not check if there is another bootstrap in progress before creating the Activiti beans that initialize the tables;
7, Obviously the second node would fail too boot because of these errors

Do you see any other sort of flagging/locking that would ensure that only one of the cluster node would do the DB initialization/upgrade without knowing IF there is even any other cluster member out there ?

andrei-rebegea:
All the logging added is with the INFO level because:
1. that is the default logging level for this package
2. this code ONLY runs at start up, so it is not like it will fill up the logs constantly
3. these kind of boot up synchronization problems usually happen sometimes and only for some cluster members
4. these kind of problems are more likely to happen with an empty DB (forget about having runtime settings for the log4j)
5. we will usually run into start up problems while working with kubernetes/helm (and docker-compose)
Imagine how "EASY" is to change the log level in a docker image! Do you want to change the log level to debug for this package (this means creating a new docker image, and redeploying or updating the deployment) when you sometimes have this problem on an empty DB?

I will stick to my decision to keep the INFO log level.
I can change some messages to be more clear, but as is now, the messages allows a developer to track down the exact parts of the code are executed by each cluster member and most likely figure out what happened if something goes wrong

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
andrei-rebegea: I can change this message. I run out of inspiration apparently

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
andrei-rebegea: see main comment below

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
killerboot: It looks like `alf_bootstrap_lock` table should exist for the whole time of the bootstrap (Alfresco and Activiti tables). The Activiti tables are always created after the Alfresco tables and that should be done by the same node. I don't think that the check for Activiti tables to exist if we just created Alfresco tables makes any sense.
I'm not sure if it is easy to refactor the code in this way. The current solution should work as well.

killerboot:
Changing the log levels in a docker image is a separate issue and is not connected to this problem.

My main point is that some INFO logs have meaningless messages like "Created schema value: false" for an Administrator or even a developer. It is required to match the log message with the source code to understand what went wrong.

andrei-rebegea:
I change the log " Created schema value: false" to 
(createdSchema ? "Created" : "Updated") + " schema in " + (System.currentTimeMillis() - start) + " ms");
and the "activiti table exists" with 
logger.info("Activiti tables need to be " + (activitiTablesExist ? "checked for patches" : " created"));

I agree that the code could be refactored to create the activiti tables right after creating the ALF tables, but since there are no tests in this area I don't fancy refactoring it.

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -857,12 +870,17 @@ private boolean updateSchema(Connection connection) throws Exception
         boolean create = false;
         try
         {
-            countAppliedPatches(connection);
+            final int numberOfPatchesApplied = countAppliedPatches(connection);
+            if (logger.isInfoEnabled())
+            {
+                logger.info("Number of patches applied detected: " + numberOfPatchesApplied);
```
derekhulley: "Applied patches detected: 5"

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -924,21 +961,40 @@ private boolean updateSchema(Connection connection) throws Exception
...
         {
+            long start = System.currentTimeMillis();
+
             // verify that all patches have been applied correctly
             checkSchemaPatchScripts(connection, preUpdateScriptPatches, false);       // check scripts
             checkSchemaPatchScripts(connection, postUpdateScriptPatches, false);      // check scripts
+
+            if (logger.isInfoEnabled())
+            {
+                logger.info("Checking all patches have been applied, executed in: " +
```
derekhulley: "Checking that all patches have been applied took 123424324ms"

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -924,21 +961,40 @@ private boolean updateSchema(Connection connection) throws Exception
                 appliedPatch.setReport("Placeholder for Activiti bootstrap at schema " + installedSchemaNumber);
                 appliedPatchDAO.createAppliedPatch(appliedPatch);
             }
+            if (logger.isInfoEnabled())
+            {
+                logger.info("Creating and patching activiti tables executed in: " + (System.currentTimeMillis() - start) + " ms");
```
derekhulley: Please use the past tense for messages after the fact:
"Creating and patching Activiti tables took 121345ms"

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -924,21 +961,40 @@ private boolean updateSchema(Connection connection) throws Exception
...
+
             // Execute any auto-update scripts for Activiti tables
             checkSchemaPatchScripts(connection, updateActivitiScriptPatches, true);
 
             // verify that all Activiti patches have been applied correctly
             checkSchemaPatchScripts(connection, updateActivitiScriptPatches, false);
+
+            if (logger.isInfoEnabled())
+            {
+                logger.info("Checking and patching activiti tables executed in: " + (System.currentTimeMillis() - start) + " ms");
```
derekhulley: "Checking and patching Activiti tables took 1234ms"

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -951,6 +1007,11 @@ private boolean updateSchema(Connection connection) throws Exception
      */
     private void initialiseActivitiDBSchema(Connection connection)
     {
+        if (logger.isInfoEnabled())
+        {
+            logger.info("Initialise Activiti DB Schema tables");
```
derekhulley: If this is before the fact, then it should be a statement of intention:
"Initialising Activiti DB schema tables."

src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java
```diff
@@ -1050,7 +1101,42 @@ else if (!apply)
...
+     * Only one node from the cluster should do the initialization of the DB, so all other
+     * nodes should just retry to execute the schema checks after the initialization of
+     * the DB has ended;
+     *
+     */
+    private void ensureCurrentClusterMemberIsBootstrapping(Connection connection) throws Exception
+    {
+        if (isAnotherClusterMemberBootstrapping(connection))
+        {
+            logger.info("Another cluster node is doing a DB schema initialization or DB update. "
```
derekhulley: "Another Alfresco cluster node is updating the DB.  Waiting for the other node to finish ..."

PR #227, opened 2018-10-25
REPO-3876: Fix Remote User Mapper

eknizat:
* Set timeout to the http client used by keycloak to retrieve a public key
* Ensure that the user mapper does not throw exceptions
* IdentityServiceRemoteUserMapper should not attempt basic authentication

### DIFF
--- a/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceDeploymentFactoryBean.java
+++ b/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceDeploymentFactoryBean.java
@@ -56,14 +56,22 @@ public class IdentityServiceDeploymentFactoryBean implements FactoryBean<Keycloa
     {
         KeycloakDeployment deployment = KeycloakDeploymentBuilder.build(this.identityServiceConfig);
 
-        // Set client with custom timeout values. This can be removed if the future versions of Keycloak accept timeout values through the config.
-        int connectionTimeout = identityServiceConfig.getClientConnectionTimeout();
-        int socketTimeout = identityServiceConfig.getClientSocketTimeout();
-        HttpClient client = new HttpClientBuilder()
-                .establishConnectionTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
-                .socketTimeout(socketTimeout, TimeUnit.MILLISECONDS)
-                .build(this.identityServiceConfig);
-        deployment.setClient(client);
+        // Set client with custom timeout values if client was created. This can be removed if the future versions of Keycloak accept timeout values through the config.
+        if (deployment.getClient() != null)
+        {
+            int connectionTimeout = identityServiceConfig.getClientConnectionTimeout();
+            int socketTimeout = identityServiceConfig.getClientSocketTimeout();
+            HttpClient client = new HttpClientBuilder()
+                    .establishConnectionTimeout(connectionTimeout, TimeUnit.MILLISECONDS)
+                    .socketTimeout(socketTimeout, TimeUnit.MILLISECONDS)
+                    .build(this.identityServiceConfig);
+            deployment.setClient(client);
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceRemoteUserMapper.java
```diff
@@ -179,28 +187,10 @@ private String extractUserFromHeader(HttpServletRequest request)
         }
         else
         {
-            // if bearer token failed, try basic auth, if enabled
-            if (this.keycloakDeployment.isEnableBasicAuth())
+            // Basic auth in IdentityServiceRemoteUserMapper was removed as part of https://issues.alfresco.com/jira/browse/REPO-3876
```
skopf: Not sure if we need this comment

src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceDeploymentFactoryBean.java
```diff
@@ -51,7 +55,14 @@ public void setIdentityServiceConfig(IdentityServiceConfig config)
     public KeycloakDeployment getObject() throws Exception
     {
         KeycloakDeployment deployment = KeycloakDeploymentBuilder.build(this.identityServiceConfig);
-        
+
+        // Set client with custom timeout values. This can be removed if the future versions of Keycloak accept timeout values through the config.
+        HttpClient client = new HttpClientBuilder()
+                .establishConnectionTimeout(2, TimeUnit.SECONDS)
+                .socketTimeout(2, TimeUnit.SECONDS)
```
skopf: 1. Are there other timeouts as well? What if we could establish the connection real quick and the TCP connection is stable, but the other end simply does not respond? There should be another timeout on the HTTP level
2. Not sure if there are situations where we need to increase this timeout. perhaps we should read these values from prop files in the end. Just to have a possibility for support to increase them in some edge case scenarios

src/test/resources/alfresco-global.properties
```diff
@@ -43,4 +43,6 @@ identity-service.enable-pkce=true
 identity-service.ignore-oauth-query-parameter=true
 identity-service.credentials.secret=11111
 identity-service.credentials.provider=secret
+identity-service.client-socket-timeout=1000
```
killerboot: Please add a comment about the units, milliseconds in this case.

src/test/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceRemoteUserMapperTest.java
```diff
@@ -85,24 +85,24 @@
...
-    private IdentityServiceConfig identityServiceConfig;
-
-    /* (non-Javadoc)
-     * @see junit.framework.TestCase#setUp()
-     */
-    @Override
-    protected void setUp() throws Exception
+    private IdentityServiceConfig identityServiceConfig;
+
+    /* (non-Javadoc)
```
killerboot: We don't need such JavaDocs any more, do we?

src/test/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceRemoteUserMapperTest.java
```diff
@@ -113,19 +113,19 @@ protected void setUp() throws Exception
...
-    protected void tearDown() throws Exception
-    {
-        childApplicationContextManager.destroy();
-        childApplicationContextManager = null;
-        childApplicationContextFactory = null;
-    }
+                    getApplicationContext().getBean(CONFIG_BEAN_NAME);
+    }
+    
+    /* (non-Javadoc)
```
killerboot: Odd JavaDoc...

PR #228, opened 2018-10-25
MNT-20139: CmisConnector returns wrong values for changeLogToken and hasMoreItems

vzorge:
3 bugs have been fixed in this pull request. For detailed description see below.

**Short description of the fix**
Ensure that the returned token is always updated. This is currently not the case when there are no more items. The next token is set to null in that case, even though there might be (maxItems-1) results. The comments in the code no longer reflected the actual code, which is fixed as well.

This causes bugs where the calling code would get say 900 of a 1000 requested items but a token of null. The only way forward was to get all the items again and ignore duplicates.

**Detailed description of the bugs**

_Issue 1: Returned token from cmis http call to CMISChangeLog returns the wrong changeLogtoken_
When a query is done with Apache Chemistry or manually calling the url it generates, the returned changeLog is always incorrect.  

To start with a call when there are more items available from the given changeLogToken.
 
​Use a local docker which has seen quite a few documents being added and moved around, so that there are changes in the changelog
Call ​http://localhost:8080/alfresco/api/-default-/public/cmis/versions/1.1/browser?cmisselector=contentChanges&changeLogToken=0&includeProperties=false&includePolicyIds=false&includeACL=false&maxItems=1000&succinct=true
In my case this returns Json:
```
{
    objects: [array of 1000 objects],
    hasMoreItems: true,
    changelogToken: "1000"
}
```
This changelogToken is always exactly the old changeLogToken + maxItems
Now call a lower level API to get these same 1000 documents: http://localhost:8080/alfresco/service/api/audit/query/CMISChangeLog?verbose=true&limit=1000&forward=true&fromId=0
The result here is: {count: 1000, entries:[array of 1000 documents]}
We can inspect the last document and see what its ID is, which is its changelogToken. In my case, this is 11550
Expected: I would have expected the changelogToken variable in step 3 to be 11550.

_Issue 2: returned changeLogToken is null when there are no more changes from the changeLog_
This is what happends there are no more items available from the given changeLogToken af the call
 

​same as before, set up an Alfresco
same call, but change the changeLogToken variable to the latest from your repository and subtract a few more from it. So for me the latest from the respository is 6317, so I would supply 6000 as my changeLogToken: http://localhost:8080/alfresco/api/-default-/public/cmis/versions/1.1/browser?cmisselector=contentChanges&changeLogToken=6000&includeProperties=false&includePolicyIds=false&includeACL=false&maxItems=1000&succinct=true
This now returns the following json:
```
{
    objects: [array of 14],
    hasMoreItems: false,
    changeLogToken: null,
}
```
Expected: the value of the latest entry as changeLogToken, not null.

_Issue 3: Changelog returns hasMoreItems = false, even though there are more changes._

This is somewhat harder to reproduce, because I don't know how to set up the Alfresco changelog data for it. I do know why it happends. This happends when an entry in the changelog, requested through the /service/api/audit/query/CMISChangeLog call has a 'values' field that is an empty object.  Example:
```
{
    id: 700,
    application:	"CMISChangeLog",
    user: "admin",
    time: "2018-10-19T08:21:14.186Z",
    values: {}
}
```
If you have an entry like the above, doing the following steps results in a problem:
 
​Made a call to Alfresco with startToken 0 and request 1000 changes: http://localhost:8080/alfresco/api/-default-/public/cmis/versions/1.1/browser?cmisselector=contentChanges&changeLogToken=0&includeProperties=false&includePolicyIds=false&includeACL=false&maxItems=100&succinct=true
Result:
```
{
    objects: [array of 100],​
    hasMoreItems: false,
    changeLogToken: null
}
```
However, if I do the lower level call: http://localhost:8080/alfresco/service/api/audit/query/CMISChangeLog?verbose=true&limit=100&forward=true&fromId=0
I do get 100 entries.
Expected​: I would epect the higher level call to know that there are more items, even though some were filtered out.
The problem here is that the CmisConnector uses the resulting documents array to determine if there are more items, even though this array is subject to filtering. This array may even be bigger then the requested value (code path supports it, I do not know if it can happen in practise).

### DIFF
--- a/src/main/java/org/alfresco/opencmis/CMISConnector.java
+++ b/src/main/java/org/alfresco/opencmis/CMISConnector.java
@@ -3675,11 +3675,15 @@ public class CMISConnector implements ApplicationContextAware, ApplicationListen
         final ObjectListImpl result = new ObjectListImpl();
         result.setObjects(new ArrayList<ObjectData>());
 
+        // Collect entryIds to use a counter and a way to find the last changeLogToken
+        final List<Long> entryIds = new ArrayList<Long>();
+
         EntryIdCallback changeLogCollectingCallback = new EntryIdCallback(true)
         {
             @Override
             public boolean handleAuditEntry(Long entryId, String user, long time, Map<String, Serializable> values)
             {
+                entryIds.add(entryId);
                 result.getObjects().addAll(createChangeEvents(time, values));
                 return super.handleAuditEntry(entryId, user, time, values);
             }
@@ -3694,7 +3698,7 @@ public class CMISConnector implements ApplicationContextAware, ApplicationListen
             }
             catch (NumberFormatException e)
             {
-                throw new CmisInvalidArgumentException("Invalid change log token: " + changeLogToken);
+                throw new CmisInvalidArgumentException("Invalid change log token: " + changeLogToken.getValue());
             }
         }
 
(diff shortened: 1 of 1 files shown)


vzorge:
I have raised this issue with our Alfresco supplier and they were going to contact Alfresco support and log an issue. When I get their feedback, I will update this PR. I don't want to create duplicate issues if I can help it.
I already submitted this pull request so that you supplier could more easily log the issue.

vzorge:
I created an issue (because our supplied had not done so yet and asked me to). https://issues.alfresco.com/jira/browse/ALF-22043

I did not make any tests for two reasons:
1. There were only 2 tests on this method to begin with
2. I am a developer of an application that uses Alfresco for data storage and the company I work for is the user that is impacted by this bug. Our supplier could not find this issue themselves, so I took matters into my own hands, but I can not justify spending more time on this fix then I have already done.

derekhulley:
Thanks.  Yes, your supplier should have asked for a Support issue for the customer, assuming the customer is an Alfresco customer.  For the Community version, ALF issues are required.
Please email me directly (derek dot hulley at Alfresco dot com) and provide the customer details and we can check.

PR #236, opened 2018-10-26
REPO-3944 Renditions: Disabling the local transform switch should not  disable the transform server

alandavis:
The ThumbnailService was not taking the newer RenditionService2 into account.


alandavis:
Also includes changes to TransformerDebug so we can see TransformerService requests and responses in the log.

alandavis:
https://bamboo.alfresco.com/bamboo/browse/PLAT-REP224-17 Green Build

PR #242, opened 2018-10-29
REPO-3972  doc to pdf is not sent to the Transform Service

alandavis:
The RemoteTransformClient was not sending the transform to the Transform Service because, the "pdf" rendition had an extra allowEnlargement option, which had been added in 6.1 EA3 as a way to force the conversion to use the ImageTransformationOptions sub class rather then TransformationOptions. While investigation this bug, it was found the type should have been SWFTransformationOptions, which had been incorrectly added in ACS 5.2

Green build: https://bamboo.alfresco.com/bamboo/browse/PLAT-REP263-UNIT-2

### DIFF
--- a/src/main/java/org/alfresco/repo/rendition2/TransformationOptionsConverter.java
+++ b/src/main/java/org/alfresco/repo/rendition2/TransformationOptionsConverter.java
@@ -112,7 +112,7 @@ public class TransformationOptionsConverter implements InitializingBean
 
     private static Set<String> PDF_OPTIONS = new HashSet<>(Arrays.asList(new String[]
             {
-                    PAGE, WIDTH, HEIGHT, ALLOW_ENLARGEMENT, MAINTAIN_ASPECT_RATIO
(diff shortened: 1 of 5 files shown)


src/main/java/org/alfresco/repo/rendition2/TransformationOptionsConverter.java
```diff
@@ -259,12 +264,19 @@ else if (IMAGE_OPTIONS.containsAll(subclassOptionNames) ||  PDF_OPTIONS.contains
                 }
             }
         }
+        else if (!hasOptions)
```
cristianturlica: If isPdfRendition is true then it will not execute this code, is that the intended behaviour? The comment states that it should be used by "pdf" rendition...

alandavis: Thanks, you are correct that there is something wrong here. I think the 'if check' is redundant. 

Re: comment about pdf, also see comment on line 185

PR #263, opened 2018-11-02
APPSREPO-549 : [3.1]CLONE - Error when Declare Record Rule has "Hide" selected

constantinpopa10:
- small fix to prevent IndexOutOfBoundsException when generating RecordCreatedEvent for records that have been hidden on creation


constantinpopa10:
A test has been added on the sync AMP codebase:
See test **test007GenerateRecordCreatedEvent_WithoutSecondaryAssociations()** in file
https://git.alfresco.com/alfresco-services/dsync-services/blob/master/alfresco-device-sync-repo/src/test/java/org/alfresco/repo/events/SyncEventsServiceImplTest.java

constantinpopa10:
A fix for this issue already exists in the Sync service amp, but a fix also needs to exist in repository, for the cases when ACS runs with RM but without the Sync  amp.

PR #271, opened 2018-11-07
MNT-20139: CmisConnector returns wrong values for changeLogToken and hasMoreItems

aforascu:
- ensure that the returned token is always updated, also when there are no more items. The next token was set to null in that case, even though there might be (maxItems-1) results.
- added Junit


derekhulley:
@vzorge this is going into the code line today.

PR #272, opened 2018-11-08
Feature/REPO: 3882 Support username password authentication

eknizat:
* Implement new authentication component which authenticates users by obtaining an access token using user's credentials 
* Add keycloak-authz-client dependecy

### DIFF
--- a/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponent.java
+++ b/src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponent.java
@@ -28,40 +28,44 @@ package org.alfresco.repo.security.authentication.identityservice;
 import org.alfresco.repo.management.subsystems.ActivateableBean;
 import org.alfresco.repo.security.authentication.AbstractAuthenticationComponent;
 import org.alfresco.repo.security.authentication.AuthenticationException;
-import org.alfresco.util.PropertyCheck;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.keycloak.authorization.client.AuthzClient;
 import org.keycloak.authorization.client.util.HttpResponseException;
-import org.springframework.beans.factory.InitializingBean;
 
 /**
  *
  * Authenticates a user against Keycloak.
  * Keycloak's {@link AuthzClient} is used to retrieve an access token for the provided user credentials,
  * user is set as the current user if the user's access token can be obtained.
+ * <br>
+ * The AuthzClient can be null in which case this authenticator will just fall through to the next one in the chain.
  *
  */
-public class IdentityServiceAuthenticationComponent extends AbstractAuthenticationComponent implements InitializingBean,
-        ActivateableBean
+public class IdentityServiceAuthenticationComponent extends AbstractAuthenticationComponent implements ActivateableBean
 {
     private final Log logger = LogFactory.getLog(IdentityServiceAuthenticationComponent.class);
     private AuthzClient authzClient;
     private boolean active;
 
-    public void setKeycloakAuthzClient(AuthzClient keycloakAuthzClient)
+    public void setAuthenticatorAuthzClient(AuthzClient authenticatorAuthzClient)
     {
-        this.authzClient = keycloakAuthzClient;
-    }
-
-    @Override
-    public void afterPropertiesSet() throws Exception
-    {
-        PropertyCheck.mandatory(this, "authzClient", authzClient);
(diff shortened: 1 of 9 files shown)


src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponent.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public boolean isActive()
+    {
+        return active;
+    }
+
+    @Override
+    protected boolean implementationAllowsGuestLogin()
+    {
+        return true;
```
killerboot: We might need to have this configurable.
Guest logins are configurable in other auth subsystems.

src/main/resources/alfresco/subsystems/Authentication/identity-service/identity-service-authentication-context.xml
```diff
@@ -18,6 +15,21 @@
...
+      <property name="authenticatorAuthzClient">
+         <ref bean="authenticatorAuthzClient"/>
+      </property>
+   </bean>
+
+   <bean name="authenticatorAuthzClient" class="org.alfresco.repo.security.authentication.identityservice.AuthenticatorAuthzClientFactoryBean">
+      <property name="identityServiceConfig">
+         <ref bean="identityServiceConfig" />
+      </property>
+      <property name="disabled">
```
killerboot: It is minor, but I think it is better to have a usual "enabled" flag as opposing to "disabled" flag.

src/test/java/org/alfresco/repo/security/SecurityTestSuite.java
```diff
@@ -102,6 +103,7 @@ public static Test suite()
 
 		suite.addTestSuite(DefaultRemoteUserMapperTest.class);
 		suite.addTestSuite(IdentityServiceRemoteUserMapperTest.class);
+        suite.addTest(new JUnit4TestAdapter(IdentityServiceAuthenticationComponentTest.class));
```
killerboot: Does it need extra whitespaces?

src/test/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponentTest.java
```diff
@@ -0,0 +1,105 @@
...
+    private IdentityServiceAuthenticationComponent authComponent = new IdentityServiceAuthenticationComponent();
+    AuthenticationContext authenticationContext;
+
+    @Mock
+    AuthzClient mockAuthzClient;
+
+    @Before
+    public void setUp()
+    {
+
```
killerboot: Whiteline not required

src/test/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponentTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.context.ApplicationContext;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class IdentityServiceAuthenticationComponentTest
+{
+
+    ApplicationContext ctx = ApplicationContextHelper.getApplicationContext();
```
killerboot: Can you please use org.alfresco.util.BaseSpringTest as a parent instead.
I think we should avoid using ApplicationContextHelper, see my suggested #181

src/main/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponent.java
eknizat: I think it would still be configurable as the abstract class has the same flag which takes precedence over the implementation flag. Not sure why that this is actually there but I added it anyway in case it is used somewhere. I'v also added it into the subsystem properties.

src/test/java/org/alfresco/repo/security/authentication/identityservice/IdentityServiceAuthenticationComponentTest.java
```diff
@@ -0,0 +1,123 @@
...
+    @After
+    public void tearDown()
+    {
+        authenticationContext.clearCurrentSecurityContext();
+    }
+
+    @Test ( expected=AuthenticationException.class)
+    public void testAuthenticationFail()
+    {
+
```
andrei-rebegea: empty line. my view is that it is not required

PR #264, opened 2018-11-02
REPO-3743: Prevent the creation of built-in authorities, like user SYSTEM, GUEST

anechifor:
…STEM, GUEST.

PR #276, opened 2018-11-19
REPO-3743: Prevent the creation of built-in authorities, like user SYSTEM, GUEST

anechifor:
See :https://github.com/Alfresco/alfresco-repository/pull/264. 
Are same modification but more clear.


src/main/java/org/alfresco/repo/security/person/PersonServiceImpl.java
```diff
@@ -523,6 +523,11 @@ else if (autoCreateHomeFolderAndMissingPersonIfAllowed)
      */
     public boolean personExists(String caseSensitiveUserName)
     {
+        if(isBuiltInAuthorities(caseSensitiveUserName))
```
cristianturlica: code format required

PR #281, opened 2018-11-21
[ MNT-19763 ] Sorting on smart folders using custom attributes fails

agazzarini:
The PR contains: 

- a change in the SortConstraint class for using the prefixed attribute name (i.e. prefix:localname) instead of local name (the namespace was completely ignored) 
- the corresponding fix to the unit tests

The proposed solution doesn't include an integration test because at the moment we don't have the required infrastructure for setting up (through API) the required environment with the steps described in the issue.


meenalbhave:
@michaelsuzukisagi , as spoken to @agazzarini , the fix is in the code that is (executed) specifically for smart folders only and the sorting doclib items in share works without the fix too.
Since our test projects can not start the test environment with smart folders enabled, it doesn't seem possible in the current setup to get the integration test working.

PR #291, opened 2018-12-05
REPO-4082 Transformers should have unique option names

alandavis:


alandavis:
In addition to making the transform options unique, the following PR also includes some tidy up activities:
* Rendition 2 Definitions beans have been simplified so there is just one common parent,
  as this makes it clearer what the transform options are even if some of them are shared. 
* Removal of "required": "false" from transform-service-config.json as this is the default.
  Test added to make sure it really is the default for option values and option groups.
* Refactored some of the code in RenditionTest and a sub classes (in alfresco-enterprise-repository)
  now that the Transform Service is more stable. They are now able to share static finals.

PR #292, opened 2018-12-06
REPO-4002: Upgrade mariadb driver

eknizat:
* Upgrade mariadb driver to 2.3.0
* Remove mariadb dialect and schema which are no longer needed as they are now the same as mysql


derekhulley:
Neat!

PR #299, opened 2018-12-12
Fix/MNT-20134 label ordering not working

eliaporciani:


src/test/java/org/alfresco/repo/jscript/ScriptSearchTest.java
```diff
@@ -0,0 +1,168 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
```
tunaaksoy: Would have been good to update the date to 2018 but it's not a big deal

PR #302, opened 2018-12-17
Add Travis config

killerboot:
An example of build plan can be seen here: https://travis-ci.org/killerboot/alfresco-repository

The tests were refactored to run Spring annotations instead of ApplicationContextHelper to mitigate the problems of Camel Context loading. The refactored tests are all part of AppContext06TestSuite.

Disabled PdfBoxContentTransformerTest as it requires installation of additional software on the build agent. The test is deprecated already.


src/test/java/org/alfresco/AppContext06TestSuite.java
```diff
@@ -35,10 +35,23 @@
  * Repository project tests using the main context alfresco/application-context.xml.
  * To balance test jobs tests using this context have been split into multiple test suites.
  * Tests marked as DBTests are automatically excluded and are run as part of {@link AllDBTestsTestSuite}.
+ *
+ * <p>
+ *    All fo the tests are using Spring annotations to load full application context, see BaseSpringTest
```
derekhulley: Typo: "All of"

killerboot: Thanks, corrected.

PR #305, opened 2018-12-19
REPO-4090: mariadb queries changes.

anechifor:
REPO-4090: The backslash character \ as an escape character within strings is the default and no escape is needed. This is valid for mariadb and Mysql.


killerboot:
It looks like the code for escape characters goes back to ALF-6808 and refers to SQL 92 standard.

eknizat:
I agree that MySQL dialect should not be changed if there are no problems with MySQL.
If this is the approach to take then a MariaDB dialect can be added very easily.

I removed the MariaDB schema reference files as they are now the same as MySQL. The MariaDB dialect was removed as well as it was only added because of those schema reference files in the first place.
But if you need, you can refer to the commit below and add just the dialect part back in.

Commit removing the reference files with MariaDD dialect:
https://github.com/Alfresco/alfresco-repository/commit/694dc5c4ca07a8e899d8da94303e8f564e630aee#diff-936ed426ee53e5a35fc2f18b114d8948.

PR #312, opened 2019-01-07
REPO-4132: After upgrade the repository is empty

ancutam:
The source of the bug is a change in the `PersonServiceImpl.personExists` method which considers now built-in authorities (system, guest) as existing by default.

The System case was dealt with after checking on the person existence, so I switched between the two `ifs`. 
I checked for similar usages in the code and it doesn't seem to be any other left.

### DIFF
--- a/src/main/java/org/alfresco/repo/jscript/app/UsernamePropertyDecorator.java
+++ b/src/main/java/org/alfresco/repo/jscript/app/UsernamePropertyDecorator.java
@@ -2,7 +2,7 @@
  * #%L
  * Alfresco Repository
  * %%
- * Copyright (C) 2005 - 2016 Alfresco Software Limited
+ * Copyright (C) 2005 - 2019 Alfresco Software Limited
  * %%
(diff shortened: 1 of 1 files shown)


killerboot:
Shouldn't we add any tests? :)

ancutam:
@killerboot I didn't find a proper location for the tests, since this is caused by a _misuse of the new code_, but I guess I can try to add a test to check if the username property decoration is successful ...

ancutam:
I tried adding a test, but to validate the fix, I need a node created by the user 'System', which is no longer possible to do, since admin is the underlying user for System.
I tried to manually change the creator to System, but that is not possible, as the property `creator` is considered an auditable property, so it can't be changed.

killerboot:
Is it possible to add a unit test that will construct the UsernamePropertyDecorator with mocked personService and nodeService? If so we don't need to bother with real nodes and still test the #decorate method even if the personService#personExists returns true for "System" now. Although the value of such test would be less than an integration.

PR #319, opened 2019-01-10
ALF-22056: onCopyComplete-Policy not called in order of copy-action

jkluttig:
Changed implementation of map so that the insertion order is preserved, with this the onCopyComplete-Policy is called in the right order in case of an recursive copy.

The LinkedHashMap preserves the insertion order of the entries, so the Policy is called in the exact order of copy-operations.

I did not know how to extend CopyServiceImplTest for this, if anyone has an idea, let me know. Otherwise i would say that the change will have no effect to other components.

### DIFF
--- a/src/main/java/org/alfresco/repo/copy/CopyServiceImpl.java
+++ b/src/main/java/org/alfresco/repo/copy/CopyServiceImpl.java
@@ -31,6 +31,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


andrei-zapodeanu-alfresco:
Hello,

Can you please provide an actual example of the ids that are not returned correctly?
What is the expected behavior and what results are you getting? 

Thank you

jkluttig:
Hello,

situation:
We have a business number which identifies all documents within a folder.
-folder
--documentA -> 1
--documentB -> 2
--documentC -> 3
The number generation is implemented with the onCreateNode-Policy. When copying a folder our business case assumes that the business number of the documents is the same as in the source, but unfortunatly it is not. For example:
-copy of folder
--documentA -> 1
--documentB -> 3
--documentC -> 2
This is because of the unordered call of the copied-children within the HashMap. If the copied-children are added to the LinkedList, then the order is correct.

PR #321, opened 2019-01-11
Fix/mnt 20134/6.56.n

eliaporciani:
Backporting of changes about MNT-20134 on support/SP/6.56.N branch


tunaaksoy:
Can you fix your IDE so that you do not change the line endings, whitespace removal, etc. please? For two lines of code change half of the class has been marked as changed. The back port looks good to me.

PR #323, opened 2019-01-16
ACE-5943 / REPO-4132: Rejecting a record in RM doesn't work with 6.1.0-RC4

ancutam:
The reject action in RM triggers an email notification from repo. Code in `EMailNotificationProvider.buildTemplateModel(...)` requests for the person associated with the current authenticated user.
Potentially another bug in `Repository.getPerson()` (next call in the stack trace), considers the current user to be `AuthenticationUtil.getRunAsUser()`, instead of the fully authenticated one. This calls, down the line, for the retrieval of the person for the username `System`, which ends up in trying to actually create the user System (which is not permitted), hence the exception.

The most simple (and safe) fix was to add a check for username System, in the low-level method `PersonServiceImpl.getPersonImpl(...)`, and return `null`.

2 tests were added, for the Person service and for the email notification generation.

### DIFF
--- a/src/test/java/org/alfresco/repo/security/person/PersonTest.java
+++ b/src/test/java/org/alfresco/repo/security/person/PersonTest.java
@@ -47,6 +47,7 @@ import org.alfresco.query.PagingRequest;
 import org.alfresco.query.PagingResults;
 import org.alfresco.repo.policy.BehaviourFilter;
 import org.alfresco.repo.security.authentication.AuthenticationUtil;
+import org.alfresco.repo.security.authentication.AuthenticationUtil.RunAsWork;
 import org.alfresco.repo.security.authentication.MutableAuthenticationDao;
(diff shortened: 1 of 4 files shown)


src/test/java/org/alfresco/repo/security/person/PersonTest.java
```diff
@@ -1788,4 +1789,20 @@ public void testPreventChangesToOtherUsersPropertiesWithoutHavingAdminRights()
...
+    
+    public void testBuitInSystemUser()
+    {
+
+        AuthenticationUtil.runAsSystem(new RunAsWork<Void>()
+        {
+            @Override
+            public Void doWork()
+            {
+                personService.getPerson(AuthenticationUtil.SYSTEM_USER_NAME);
```
killerboot: Do we need to check if this returns `null`?

alandavis: Now throws an Exception

PR #329, opened 2019-01-18
ALF-22060 Fix Reader on the backing store is obtained twice in CachingContentStore

nbarithel:
Fix and unit test for https://issues.alfresco.com/jira/browse/ALF-22060

### DIFF
--- a/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java
+++ b/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java
@@ -206,7 +206,7 @@ public class CachingContentStore implements ContentStore, ApplicationEventPublis
                     return backingStoreReader;
                 }
                 
-                ContentReader reader = attemptCacheAndRead(url);
+                ContentReader reader = attemptCacheAndRead(url, backingStoreReader);
                 
                 if (reader != null)
                 {
@@ -254,14 +254,14 @@ public class CachingContentStore implements ContentStore, ApplicationEventPublis
      * @param url URL to cache.
      * @return A reader onto the cached content file or null if unable to provide one.
      */
-    private ContentReader attemptCacheAndRead(String url)
+    private ContentReader attemptCacheAndRead(String url, ContentReader backingStoreReader)
     {
         ContentReader reader = null;
         try
         {
             if (!cache.contains(url))
             {
-                if (cache.put(url, backingStore.getReader(url)))
+                if (cache.put(url, backingStoreReader))
                 {
                     reader = cache.getReader(url);
                 }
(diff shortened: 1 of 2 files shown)


nbarithel:
Hi @skopf,

I am not sure that you understood this PR correctly. I am well aware of the difference between the cachingStore ContentReader and the backingStore ContentReader. This PR is only about reusing the backingStore ContentReader. It is useful especially with object storage where size and content do retrieve object's metadata first to build the ContentReader.

I agree with your detailed description, what you are saying is true, but my PR is not about that.

Note: if you look at the diff, it is mainly replacing backingStore.getReader(url) with backingStoreReader, nothing else noteworthy, the rest is boilerplate.

Going into details:
1. First, we agree that [CachingContentStore.java#L201](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L201) is just to retrieve the size from the backingStore, without retrieving the content.
2. Then, in fact in [CachingContentStore.java#L209](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L209) it is for retrieving the content from the backingStore if and only if the cache does not contain the url:
[CachingContentStore.java#L262](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L262)
3. To do that, it is getting yet another backingStore reader [CachingContentStore.java#L264](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L264), although it already took one at [CachingContentStore.java#L201](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L201)
4. You mentioned [ContentCacheImpl.java#L131](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/ContentCacheImpl.java#L131) but it is in fact in the put before that in [ContentCacheImpl.java#L146](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/ContentCacheImpl.java#L146) that the content is copied from the backingStore to a temp file so that next call to the getReader will return a reader on this file! 
5. So this PR is just reusing the backingStore ContentReader used to retrieve the size, to retrieve the content if needed, I do not see what can go wrong with that. In every case, [CachingContentStore.java#L271](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L271) (that you mentioned) and [CachingContentStore.java#L266](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/content/caching/CachingContentStore.java#L266) it returns the cachingStore ContentReader.

HTH

Nicolas

PR #336, opened 2019-01-30
REPO-3962: Remove context refresh from testOpenCloseOpenCloseFull

eknizat:
* The refreshing of our core application context does not work with Spring version greater than 5.1.0
* We don't need to test refreshing the core application context as that never happens, only the subsystem application contexts are refreshed

### DIFF
--- a/src/test/java/org/alfresco/RepositoryStartStopTest.java
+++ b/src/test/java/org/alfresco/RepositoryStartStopTest.java
@@ -190,6 +190,31 @@ public class RepositoryStartStopTest extends TestCase
        ApplicationContextHelper.closeApplicationContext();
        assertNoCachedApplicationContext();
     }
+
+    /**
+     *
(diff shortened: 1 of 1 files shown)


src/test/java/org/alfresco/RepositoryStartStopTest.java
```diff
@@ -187,9 +186,6 @@ public void ignoredTestOpenCloseOpenCloseFull() throws Exception
        ctx = getFullContext();
        assertEquals(ctx, ctx2);
        
-       // Refresh it, shouldn't break anything
-       ((AbstractApplicationContext)ctx).refresh();
```
andrei-rebegea: Erik, I would rather prefer to keep these lines here - commented
and also add a java comment linking to https://issues.alfresco.com/jira/browse/REPO-4176 
I also think that enabling this context refresh should be part of the acceptance criteria on that repo ticket.

PR #337, opened 2019-01-31
REPO-4102: MNT-20213: Unable to run using Java 8

cristianturlica:
- explicit cast for compatibility with covariant return type on JDK 9's ByteBuffer


cristianturlica:
If it can compile with 11 and run with 8, then yes it would help.

killerboot:
> If it can compile with 11 and run with 8, then yes it would help.

It is possible but require a bit more configuration. You can try jdk_switcher https://docs.travis-ci.com/user/languages/java/

PR #353, opened 2019-02-26
[SEARCH-1272] Set permissionService protected to be used by Extended …

eliaporciani:
I needed to modify the modifier of permissionService field to be used in the subclasses defined in AGS.
Is it better to modify all the modifiers of the various Services from private to protected for future needs?


tpage-alfresco:
If you just want to use it then can't you do that with a getter?

eliaporciani:
Yes, it is another way. In this way the scope of permissionService will remain within the subclasses. If you think it is better I can set it back to private and add a (public?) getter.

tpage-alfresco:
I think that would be slightly better.  If you want to add other getters at the same time then that's great.

PR #356, opened 2019-02-27
REPO-4180: Service Pack: MNT-20305 Oracle schema validation check fai…

cristianturlica:
…lure with ojdbc7.jar version 12.1.0.2

   - fixed regression in SchemaBootstrap that cause the schema validation to not be done
   - added support for custom sequences retrieval (active if only schema reference sequences file exists)

### DIFF
--- a/src/main/java/org/alfresco/util/ScriptUtil.java
+++ b/src/main/java/org/alfresco/util/ScriptUtil.java
@@ -1,0 +1,107 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2019 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software.
+ * If the software was purchased under a paid Alfresco license, the terms of
+ * the paid license agreement will prevail.  Otherwise, the software is
+ * provided under the following open source license terms:
+ *
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.util;
+
+import java.io.IOException;
+import java.io.LineNumberReader;
+
+import org.springframework.core.io.support.EncodedResource;
+
+public abstract class ScriptUtil
+{
+    private static final String DEFAULT_SCRIPT_COMMENT_PREFIX = "--";
+
+    /**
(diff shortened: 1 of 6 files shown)


src/main/java/org/alfresco/util/ScriptUtil.java
```diff
@@ -0,0 +1,107 @@
...
+ * #L%
+ */
+package org.alfresco.util;
+
+import java.io.IOException;
+import java.io.LineNumberReader;
+
+import org.springframework.core.io.support.EncodedResource;
+
+public abstract class ScriptUtil
```
killerboot: I think it will be better to rename this to "DBScriptUtil" or similar to highlight this is about SQL scripts as we also JavaScript and web script.

src/main/java/org/alfresco/util/ScriptUtil.java
```diff
@@ -0,0 +1,107 @@
...
+     *            "--")
+     * @return a String containing the script lines
+     */
+    private static String readScript(LineNumberReader lineNumberReader, String lineCommentPrefix) throws IOException
+    {
+        String statement = lineNumberReader.readLine();
+        StringBuilder scriptBuilder = new StringBuilder();
+        while (statement != null)
+        {
+            if (lineCommentPrefix != null && !statement.startsWith(lineCommentPrefix))
```
killerboot: It might be better to either extract and reuse the code from SchemaBootstrap#executeScriptFile (it already handles comments, variables, loops and etc) or just remove the comment handling from here at all to avoid duplication.

src/main/java/org/alfresco/util/ScriptUtil.java
```diff
@@ -0,0 +1,107 @@
...
+     *            the resource (potentially associated with a specific encoding) to
+     *            load the SQL script from
+     * @param lineCommentPrefix
+     *            the prefix that identifies comments in the SQL script (typically
+     *            "--")
+     * @return a String containing the script lines
+     */
+    private static String readScript(EncodedResource resource, String lineCommentPrefix) throws IOException
+    {
+        LineNumberReader lineNumberReader = new LineNumberReader(resource.getReader());
```
killerboot: Looking at https://github.com/Alfresco/alfresco-repository/blob/5e42f708be424de28864051917c53a48c8eada6c/src/main/java/org/alfresco/repo/domain/schema/SchemaBootstrap.java#L1268 we might need to care about UTF-8 encoding.

src/main/java/org/alfresco/util/ScriptUtil.java
cristianturlica: Good idea, I will update it.

src/main/java/org/alfresco/util/ScriptUtil.java
cristianturlica: Yes, I updated retrieveAndProcessSequences to use UTF-8:
`final String script = DBScriptUtil.readScript(new EncodedResource(resource, "UTF-8"));`

src/main/java/org/alfresco/util/ScriptUtil.java
cristianturlica: The SchemaBootstrap#executeScriptFile provides extensive functionality and is tightly coupled. I wouldn't add that part because this would allow users to do a lot more that they should with this file.

killerboot: ok, fair enough.

PR #364, opened 2019-03-05
Mail action executor better html detection

Gagravarr:
In order to get the correct rendering in Outlook, we need to update some of our HTML email templates to include a doctype at the start. However, this causes Alfresco to send them as Plain Text instead of HTML, because the logic in `MailActionExecutor` to decide if a `FTL` template rendered into HTML or Plain Text only considers the opening HTML tag, not a HTML Doctype declaration

The logic in the `isHTML` method needs to consider this too. In addition, unit testing is required for all of the HTML detection logic, including this new check.

Also raised as ALF-22073


andrei-rebegea:
associated with https://issues.alfresco.com/jira/browse/REPO-4268 and https://issues.alfresco.com/jira/browse/ALF-22073

PR #365, opened 2019-03-06
REPO-3932: Use a fully distributed cache for user authority cache

andrei-rebegea:
Using a fully distributed cache for the user authorities used in  https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java#L128 will resolve the problems related to data about user authorities getting desynchronized across cluster nodes;
There is a complex system of distributed caches, local update triggered: AuthorityBridgeTableAsynchronouslyRefreshedCache and local DB caches like ParentAssocsCache  https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/domain/node/AbstractNodeDAOImpl.java#L195 
that just don't manage to keep up with the invalidating cache type used before for the userAuthorityCache.
We could honestly say that we barely scratched the surface with this 4 weeks investigation, but the data proves the fix works. The only down side may be for large deployments more hazelcast traffic but we think it is worth if considering the implications about data consistency.
The cache type is a property that the clients can change if they are willing to live with the consequences and they don't experience any problems with their work flows.


andrei-rebegea:
Had a review with Erik and Martin online. I will merge these changes

PR #378, opened 2019-03-26
Simplification of the code after the review.

aborroy:


src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
```diff
@@ -1,28 +1,28 @@
...
+ * 
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
```
tpage-alfresco: Are you changing the line endings deliberately?

src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
```diff
@@ -59,7 +60,7 @@
     private int index;
 
     private Map<String, Float> scores;
-
+
```
tpage-alfresco: Random extra whitespace

src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
```diff
@@ -176,7 +177,11 @@ public Serializable getValue(String columnName)
         context.setScore(getScore());
         for (Column column : query.getColumns())
         {
-            if (column.getAlias().equals(columnName))
+            // When an SCORE selector is included, score must be adapted to range 0..1 due to CMIS specification
+            if (column.getFunction().getName().equals(Score.NAME)) {
```
tpage-alfresco: Braces should be on their own line.

tpage-alfresco: Could we have an NPE here?  E.g. If a column isn't a function?

src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
```diff
@@ -221,6 +226,16 @@ public Serializable getValue(String columnName)
...
         return null;
     }
+    
+    /**
+     * CMIS Specification states that scoring results must be in a 0..1 range
+     * This function re-adapt the scores when any scoring field or expression is requested by the query. 
+     * It's a safe approach, as includes negative numbers and also paged requests.
+     * @return A value between 0 and 1
+     */
+    private float getNormalisedScore() {
```
tpage-alfresco: Trailing brace should be on a new line

src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
aborroy: Of course, not.

src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
aborroy: In fact, `function` can be Null:
https://github.com/Alfresco/alfresco-data-model/blob/master/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/lucene/LuceneQueryModelFactory.java#L264

It is not Null for any of our 1,000 e-2-e tests, but there is a possibility to be null.

src/main/java/org/alfresco/opencmis/search/CMISResultSetRow.java
tpage-alfresco: Ok!  I wasn't sure - some of our files have mixed line endings and it can be good to make the file uniform.

PR #381, opened 2019-03-29
Fix/search 1531 remove solrj

aborroy:
Dependency has been removed by copying SOLR HighlightParams constants to Alfresco code and by using HTTP Client invocations to request SOLR Admin & Backup API instead of SOLR Native client.

### DIFF
--- a/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
+++ b/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
@@ -25,13 +25,14 @@
  */
 package org.alfresco.repo.search.impl.lucene;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * JSON returned from SOLR API Parser
@@ -44,7 +45,7 @@ import org.json.JSONObject;
 public abstract class AbstractJSONAPIResult implements JSONAPIResult
 {
     
-    private static final Log logger = LogFactory.getLog(AbstractJSONAPIResult.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractJSONAPIResult.class);
     
     protected Long status; 
     protected Long queryTime;
@@ -73,15 +74,16 @@ public abstract class AbstractJSONAPIResult implements JSONAPIResult
     @Override
     public List<String> getCores()
     {
-        return cores;
+        return Collections.unmodifiableList(cores);
     }
     
     /* (non-Javadoc)
      * @see org.alfresco.repo.search.impl.lucene.JSONAPIResult#getCoresInfo()
      */
     @Override
-    public Map<String, Map<String, Object>> getCoresInfo() {
-        return coresInfo;
+    public Map<String, Map<String, Object>> getCoresInfo() 
+    {
+        return Collections.unmodifiableMap(coresInfo);
     }
     
     /**
@@ -91,10 +93,8 @@ public abstract class AbstractJSONAPIResult implements JSONAPIResult
      */
     protected void processJson(JSONObject json) throws JSONException
     {
-        if (logger.isDebugEnabled())
-        {
-            logger.debug("JSON response: "+json);
-        }
+        
+        LOGGER.debug("JSON response: {}", json);
         
         JSONObject responseHeader = json.getJSONObject("responseHeader");
         status = responseHeader.getLong("status");
--- a/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
+++ b/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
@@ -76,7 +76,7 @@ public abstract class AbstractJSONAPIResult implements JSONAPIResult
     @Override
     public List<String> getCores()
     {
-        return Collections.unmodifiableList(cores);
+        return (cores == null ? null : Collections.unmodifiableList(cores));
     }
     
     /* (non-Javadoc)
@@ -85,7 +85,7 @@ public abstract class AbstractJSONAPIResult implements JSONAPIResult
     @Override
     public Map<String, Map<String, Object>> getCoresInfo() 
     {
-        return Collections.unmodifiableMap(coresInfo);
+        return (coresInfo == null ? null : Collections.unmodifiableMap(coresInfo));
     }
     
     /**
--- a/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
+++ b/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
@@ -26,9 +26,11 @@
 package org.alfresco.repo.search.impl.lucene;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.slf4j.Logger;
@@ -102,6 +104,34 @@ public abstract class AbstractJSONAPIResult implements JSONAPIResult
         
         processCoresInfoJson(json);
 
+    }
+    
+    /**
+     * Creates a property-value Map from a JSON Object containing properties and values
+     * This method provides the right input for MBeans to expose the SOLR response values from the Response
+     * @param json Simple JSON Object containing only properties
+     * @return Property-value Map
+     * @throws JSONException
+     */
+    protected Map<String, Object> getPropertyValueMap(JSONObject json) throws JSONException
+    {
+        Map<String, Object> propertyValueMap = new HashMap<>();
+        JSONArray nodesPropertyNameList = json.names();
+        for (int j = 0; j < nodesPropertyNameList.length(); j++)
+        {
+            String propertyName = String.valueOf(nodesPropertyNameList.get(j));
+            Object propertyValue = json.get(propertyName);
+            if (propertyValue != JSONObject.NULL)
+            {
+                // MBeans Objects are defined as Long types, so we need casting to provide the expected type
+                if (propertyValue instanceof Integer)
+                {
+                    propertyValue = Long.valueOf((Integer) propertyValue);
+                }
+                propertyValueMap.put(propertyName, propertyValue);
+            }
+        }
+        return propertyValueMap;
     }
     
     /**
--- a/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
+++ b/src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
@@ -1,0 +1,114 @@
+/*
+ * #%L
+ * Alfresco Repository
+ * %%
+ * Copyright (C) 2005 - 2016 Alfresco Software Limited
+ * %%
+ * This file is part of the Alfresco software. 
+ * If the software was purchased under a paid Alfresco license, the terms of 
+ * the paid license agreement will prevail.  Otherwise, the software is 
+ * provided under the following open source license terms:
+ * 
+ * Alfresco is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * Alfresco is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Alfresco. If not, see <http://www.gnu.org/licenses/>.
+ * #L%
+ */
+package org.alfresco.repo.search.impl.lucene;
+
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * JSON returned from SOLR API Parser
+ * This class defines common properties and performs response header parsing.
+ * An abstract method is provided for implementers to parse Core Information.
+ *
+ * @author aborroy
+ * @since 6.2
+ */
+public abstract class AbstractJSONAPIResult implements JSONAPIResult
+{
+    
+    private static final Log logger = LogFactory.getLog(AbstractJSONAPIResult.class);
+    
+    protected Long status; 
+    protected Long queryTime;
+    protected List<String> cores;
+    protected Map<String, Map<String, Object>> coresInfo;
+    
+    /* (non-Javadoc)
+     * @see org.alfresco.repo.search.impl.lucene.JSONActionResult#getQueryTime()
+     */
+    public Long getQueryTime()
+    {
+        return queryTime;
+    }
+    
+    /* (non-Javadoc)
+     * @see org.alfresco.repo.search.impl.lucene.JSONActionResult#getStatus()
+     */
+    public Long getStatus()
+    {
+        return status;
+    }
+
+    /* (non-Javadoc)
+     * @see org.alfresco.repo.search.impl.lucene.JSONAPIResult#getCores()
+     */
+    @Override
+    public List<String> getCores()
+    {
+        return cores;
+    }
+    
+    /* (non-Javadoc)
+     * @see org.alfresco.repo.search.impl.lucene.JSONAPIResult#getCoresInfo()
+     */
+    @Override
+    public Map<String, Map<String, Object>> getCoresInfo() {
+        return coresInfo;
+    }
+    
+    /**
+     * Parses the JSON to set this Java Object values
+     * @param json JSONObject returned by SOLR API
+     * @throws JSONException
+     */
+    protected void processJson(JSONObject json) throws JSONException
+    {
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("JSON response: "+json);
+        }
+        
+        JSONObject responseHeader = json.getJSONObject("responseHeader");
+        status = responseHeader.getLong("status");
+        queryTime = responseHeader.getLong("QTime");
+        
+        processCoresInfoJson(json);
+
+    }
+    
+    /**
+     * Parses the JSON to set this Java Object values related to Core Information
+     * @param json JSONObject returned by SOLR API
+     * @throws JSONException
+     */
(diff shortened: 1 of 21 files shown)


src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionStatusResult.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * The results of executing a SOLR STATUS action
+ *
+ * @author aborroy
+ * @since 6.2
+ */
+public class SolrActionStatusResult implements JSONAPIResult
+{
+    private static final Log logger = LogFactory.getLog(SolrActionStatusResult.class);
```
tpage-alfresco: Generally I'd use SLF4J for new loggers, and put static final in all caps, like `LOGGER`.  There are some advantages of using SLF4J - mainly the lazy string construction:
```
LOGGER.debug("JSON response: {}", response);
```

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionStatusResult.java
```diff
@@ -0,0 +1,126 @@
...
+    public Long getQueryTime()
+    {
+        return this.queryTime;
+    }
+    
+    /**
+     * Name of the cores managed by SOLR
+     * @return A list with the names of the cores
+     */
+    public List<String> getCores() {
```
tpage-alfresco: [Style] Trailing brace

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionStatusResult.java
```diff
@@ -0,0 +1,126 @@
...
+        return this.cores;
```
tpage-alfresco: I don't know if it matters, but this allows editing the list.  We could use an unmodifiable list instead, e.g.:
https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableList(java.util.List)
Or an ImmutableList since it doesn't look like the list is updated:
https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableList.html

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionStatusResult.java
```diff
@@ -0,0 +1,126 @@
...
+    
+    /**
+     * Parses the JSON to set this Java Object values
+     * @param json JSONObject returned by SOLR API
+     * @throws JSONException
+     */
+    protected void processJson(JSONObject json) throws JSONException
+    {
+
+        if (logger.isDebugEnabled())
```
tpage-alfresco: We don't need this check if we use slf4j instead.

src/main/java/org/alfresco/repo/search/impl/solr/AbstractSolrAdminHTTPClient.java
```diff
@@ -0,0 +1,105 @@
...
+     * Executes an action or a command in SOLR using REST API 
+     * 
+     * @param httpClient HTTP Client to be used for the invocation
+     * @param url Complete URL of SOLR REST API Endpoint
+     * @return A JSON Object including SOLR response
+     * @throws UnsupportedEncodingException
+     * @throws IOException
+     * @throws HttpException
+     * @throws URIException
+     * @throws JSONException
```
tpage-alfresco: This seems like a lot of exceptions for the calling code to handle.  Is it reasonable for the calling code to deal with e.g. UnsupportedEncodingException?  (Maybe it is reasonable - since the client is passed in?)

src/main/java/org/alfresco/repo/search/impl/solr/SolrAdminClient.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.beans.factory.BeanFactoryAware;
+import org.springframework.beans.factory.InitializingBean;
+
+/**
+ * Solr Admin client interface
+ * Provides access to Actions and Commands in SOLR API
+ * @author aborroy
+ * @since 6.2
+ */
+public interface SolrAdminClient extends BeanFactoryAware, InitializingBean
```
tpage-alfresco: I'm confused by this - I couldn't see any classes implementing this interface.

src/main/java/org/alfresco/repo/search/impl/solr/SolrAdminClient.java
```diff
@@ -0,0 +1,92 @@
...
+        backup
+    }
+    
+    /**
+     * SOLR API Handlers (partial list)
+     */
+    public static enum HANDLER
+    {
+        replication
+    }
```
tpage-alfresco: These enums seem odd to me.  Why do we have partial lists?  Why aren't `backup` and `replication` in ALL_CAPS?  Are these also defined somewhere else?

src/main/java/org/alfresco/repo/search/impl/solr/SolrAdminClient.java
```diff
@@ -0,0 +1,92 @@
...
+    
+    /**
+     * Execute a COMMAND from the SOLR API
+     * @param core Name of the core to execute the command
+     * @param handler Name of the handler for the SOLR REST API
+     * @param command Name of the command to be invoked
+     * @param parameters Action parameters in pairs of key, value
+     * @return
+     */
+    public JSONAPIResult executeCommand(String core, HANDLER handler, COMMAND command, Map<String, String> parameters);
```
tpage-alfresco: Given you've defined some methods then there must be something implementing this interface - I just couldn't find it.  Nice clear comments btw!

src/main/java/org/alfresco/repo/solr/SOLRAdminClient.java
```diff
@@ -55,68 +61,42 @@
...
 /**
  * Provides an interface to the Solr admin APIs, used by the Alfresco Enterprise JMX layer.
  * Also tracks whether Solr is available, sending Spring events when its availability changes.
  * 
  * @since 4.0
  *
  */
-public class SOLRAdminClient implements ApplicationEventPublisherAware, DisposableBean
+public class SOLRAdminClient extends AbstractSolrAdminHTTPClient
+        implements ApplicationEventPublisherAware, DisposableBean, SolrAdminClient
```
tpage-alfresco: Here's the implementation!  It's probably not a good idea to have two classes that differ only in case: `SOLRAdminClient` and `SolrAdminClient`.  Maybe put `Interface` on the end of the interface?  Usually we'd call the implementation `Impl`, but that probably involves changing more code.

src/main/java/org/alfresco/repo/solr/SOLRAdminClient.java
```diff
@@ -55,68 +61,42 @@
...
 
 	private ApplicationEventPublisher applicationEventPublisher;
 	private SolrTracker solrTracker;
 	
-	private HttpClientFactory httpClientFactory;
     private Scheduler scheduler;
+    
+    private List<SolrStoreMapping> storeMappings;
+    
+    private HashMap<StoreRef, SolrStoreMappingWrapper> mappingLookup = new HashMap<>();
```
tpage-alfresco: We probably don't need to define this as a HashMap (e.g. will Map do?)

src/main/java/org/alfresco/repo/solr/SOLRAdminClient.java
```diff
@@ -147,76 +122,176 @@ public void setScheduler(Scheduler scheduler)
...
-	    	StringBuilder sb = new StringBuilder();
-	    	sb.append(httpClientFactory.isSSL() ? "https://" : "http://");
-	    	sb.append(solrHost);
-	    	sb.append(":");
-	    	sb.append(httpClientFactory.isSSL() ? solrSSLPort: solrPort);
-	    	sb.append(baseUrl);
-			this.solrUrl = sb.toString();
-			HttpClient httpClient = httpClientFactory.getHttpClient();
+		this.solrTracker = new SolrTracker(scheduler);
+	}
```
tpage-alfresco: Mixed tabs and spaces?  (I think it was mixed before you started)

src/main/resources/alfresco/subsystems/Search/solr6/solr-search-context.xml
```diff
@@ -169,21 +169,17 @@
...
-      <property name="solrConnectTimeout" value="${solr.solrConnectTimeout}"/>
-      <property name="httpClientFactory" ref="solrHttpClientFactory"/>
       <property name="baseUrl" value="${solr.baseUrl}"/>
       <property name="scheduler">
         <ref bean="searchSchedulerFactory" />
       </property>
+		<property name="storeMappings">
+			<ref bean="solr6.store.mappings" />
+		</property>
+		<property name="useDynamicShardRegistration" value="${solr.useDynamicShardRegistration}" />
```
tpage-alfresco: [Style] Tabs --> spaces/

src/main/java/org/alfresco/repo/search/impl/lucene/JSONAPIResultFactory.java
```diff
@@ -0,0 +1,122 @@
...
+ * SOLR JSON responses are parsed according to required Action, Command or Handler 
+ *
+ * @author aborroy
+ * @since 6.2
+ */
+public class JSONAPIResultFactory
+{
+    
+    /**
+     * SOLR CoreAdmin API Actions (partial list)
```
tpage-alfresco: Is this still a partial list?

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
```diff
@@ -0,0 +1,114 @@
...
+ * This class defines common properties and performs response header parsing.
+ * An abstract method is provided for implementers to parse Core Information.
+ *
+ * @author aborroy
+ * @since 6.2
+ */
+public abstract class AbstractJSONAPIResult implements JSONAPIResult
+{
+    
+    private static final Log logger = LogFactory.getLog(AbstractJSONAPIResult.class);
```
tpage-alfresco: Please can you update the logger to ALL_CAPS and use slf4j?

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+    
+    /**
+     * Parses the JSON to set this Java Object values
+     * @param json JSONObject returned by SOLR API
+     * @throws JSONException
+     */
+    protected void processJson(JSONObject json) throws JSONException
+    {
+        if (logger.isDebugEnabled())
```
tpage-alfresco: ...and once we're using slf4j we can remove all checks like this.

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
```diff
@@ -0,0 +1,114 @@
...
+        return status;
+    }
+
+    /* (non-Javadoc)
+     * @see org.alfresco.repo.search.impl.lucene.JSONAPIResult#getCores()
+     */
+    @Override
+    public List<String> getCores()
+    {
+        return cores;
```
tpage-alfresco: I don't know whether it's desired here, but we're exposing the option for callers to update the list (same with getCoresInfo).

src/main/java/org/alfresco/repo/search/impl/lucene/JSONAPIResultFactory.java
```diff
@@ -0,0 +1,122 @@
...
+            {
+                return new SolrActionFixResult(json);
+            }
+            case CHECK:
+            {
+                return new SolrActionCheckResult(json);
+            }
+            default:
+            {
+                throw new JSONException("Action " + action + " is not supported!");
```
tpage-alfresco: Maybe throw an AlfrescoRuntimeException instead?

src/main/java/org/alfresco/repo/search/impl/lucene/JSONAPIResultFactory.java
```diff
@@ -0,0 +1,122 @@
...
+     * @param json Result in JSON of the SOLR Action invoked
+     * @return
+     * @throws JSONException
+     */
+    public static JSONAPIResult buildActionResult(ACTION action, JSONObject json) throws JSONException
+    {
+        switch (action)
+        {
+            case STATUS:
+            {
```
tpage-alfresco: Are these braces for case statements what we usually use for switch statements? They look odd to me.

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionStatusResult.java
aborroy: Moving all loggers to SLF4J.

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
aborroy: Sure, just done!

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
aborroy: Done

src/main/java/org/alfresco/repo/search/impl/lucene/JSONAPIResultFactory.java
aborroy: Yes, it is.

src/main/java/org/alfresco/repo/search/impl/lucene/JSONAPIResultFactory.java
aborroy: *case* sentences shifted left to be in the same tab as *switch*

src/main/java/org/alfresco/repo/search/impl/solr/AbstractSolrAdminHTTPClient.java
aborroy: Agree

src/main/java/org/alfresco/repo/solr/SOLRAdminClient.java
aborroy: Probably I chose the best option having all the other class names into account, but we can change more names if required.

src/main/java/org/alfresco/repo/solr/SOLRAdminClient.java
aborroy: SolrClientUtil.extractMapping() is defined to accept a HashMap, so I'm using a HashMap to avoid casting on invocations.

src/main/java/org/alfresco/repo/solr/SOLRAdminClient.java
aborroy: This is removed code.

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
aborroy: Are you suggesting to use Collections.unmodifiableSet() wrapper?

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionStatusResult.java
aborroy: Ok, so yes, you are suggesting to use this approach. Yes, they are both read-only properties.

src/main/java/org/alfresco/repo/search/impl/solr/SolrAdminClient.java
aborroy: We are not supporting al SOLR REST API operations, so the lists are partial. The caps and the lowercase is to have the exactly name SOLR API is accepting.

aborroy: Changed constants to ALL_CAPS

src/main/java/org/alfresco/repo/search/impl/solr/SolrAdminClient.java
aborroy: Finally changed the name to **SolrAdminClientInterface** to mark a difference with the implementation **SOLRAdminClient**

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionAclReportResult.java
```diff
@@ -0,0 +1,110 @@
...
+                for (int j = 0; j < nodesPropertyNameList.length(); j++)
+                {
+                    String propertyName = String.valueOf(nodesPropertyNameList.get(j));
+                    Object propertyValue = core.get(propertyName);
+                    if (propertyValue != JSONObject.NULL)
+                    {
+                        // MBeans Objects are defined as Long types, so we need casting to provide the expected type
+                        if (propertyValue instanceof Integer)
+                        {
+                            propertyValue = Long.valueOf(propertyValue.toString());
```
tpage-alfresco: Why are we going via a string?

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionAclTxReportResult.java
```diff
@@ -0,0 +1,122 @@
...
+    
+    /* (non-Javadoc)
+     * @see org.alfresco.repo.search.impl.lucene.AbstractSolrActionAPIResult#processCoresInfoJson(org.json.JSONObject)
+     */
+    @Override
+    protected void processCoresInfoJson(JSONObject json) throws JSONException
+    {
+
+        cores = new ArrayList<>();
+        coresInfo = new HashMap<>();
```
tpage-alfresco: This pattern of updating member variables (rather than just setting them) makes me nervous.  Could there be multiple threads accessing these member variables?

src/main/java/org/alfresco/repo/search/impl/lucene/SolrActionNodeReportResult.java
```diff
@@ -0,0 +1,115 @@
...
+                        {
+                            // MBeans Objects are defined as Long types, so we need casting to provide the expected type
+                            if (propertyValue instanceof Integer)
+                            {
+                                propertyValue = Long.valueOf(propertyValue.toString());
+                            }
+                            coreInfo.put(propertyName, propertyValue);
+                        }
+                    }
+                    coresInfo.put(coreName, coreInfo);
```
tpage-alfresco: I feel like I've read this code already. Can this be refactored into a helper method?

aborroy: Refactorized!

src/main/java/org/alfresco/repo/search/impl/lucene/AbstractJSONAPIResult.java
```diff
@@ -76,7 +76,7 @@ public Long getStatus()
     @Override
     public List<String> getCores()
     {
-        return Collections.unmodifiableList(cores);
+        return (cores == null ? null : Collections.unmodifiableList(cores));
```
tpage-alfresco: Nice spot!

PR #388, opened 2019-04-01
REPO-4284: MNT-20212:  Download as Zip temp content never gets cleaned up by Download Cleaner

andrei-rebegea:
Modified the SystemNodeUtils to run as system when access to the "sys:download" folder is required. 
This will prevent other "sys:download" (with differnt GUID) to be created, as it was the case in the past for non admin users;
I also added code to the DownloadService and DownloadStorage classes to handle the deletion of the old download zip files that were not deleted by the DownloadsCleanupJob because of this bug.
I added support for batching for this deletion, since old repositories may potentially contain a lot of these old files that were not deleted for years.


anechifor:
Looks good to me.

PR #390, opened 2019-04-03
REPO-3452: Housekeeping: Check activiti config files

andrei-rebegea:
deleted 2 files that were not used: activiti.mysql.create.sql and adhoc.bpmn20.xml
and moved a file to the test code, since it was only used for tests: ...repo/workflow/activiti/activiti.cfg.xml → src/test/resources/activiti/activiti.cfg.xml


eknizat:
LGTM

PR #425, opened 2019-04-25
REPO-4296 Update docker and helm charts to use new ActiveMQ images

andrei-zapodeanu-alfresco:


cristianturlica:
Changes look good to me.

PR #475, opened 2019-06-07
REPO-4331: Add remaining transformers

eknizat:


src/main/resources/alfresco/repository.properties
```diff
@@ -623,13 +623,18 @@ localTransformer.pdfrenderer.url=http://localhost:8090/
 localTransformer.imagemagick.url=http://localhost:8091/
 localTransformer.libreoffice.url=http://localhost:8092/
 localTransformer.tika.url=http://localhost:8093/
+localTransformer.misc.url=http://localhost:8094/
+
+#TODO remove localTransformer.html.url
```
alandavis: Don't think you are using this any more, so it can be removed.

src/main/resources/alfresco/transformers/0140-misc.json
```diff
@@ -0,0 +1,65 @@
+{
```
alandavis: This file is no longer needed if the new misc T-Engine replies with the same value from .../transform/config, so it should be removed.

PR #499, opened 2019-06-24
REPO-4305: jmx dump password redaction inconsistent

PunchedChimera:
Added new methods to parse the InputArguments section of the JMXDump, in order to redact any instances of plain text passwords.

### DIFF
--- a/src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
+++ b/src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
@@ -220,7 +220,7 @@ public class JmxDumpUtil
         String replacement = JmxDumpUtil.PROTECTED_VALUE;
         String output = input.replaceAll(pattern, replacement);
 
-        if(output.equals(input)) //If no string have been replaced it's possible password= could be end of sprint
+        if(output.equals(input)) //If nothing has been replaced it's possible password= could be end of string
         {
             // pattern to see if the string ends with "password=", "token=" or "pwd="
             Pattern passwordAtEnd = Pattern.compile("((password=)|(token=)|(pwd=))$");
--- a/src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
+++ b/src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
@@ -30,13 +30,17 @@ import java.io.PrintWriter;
 import java.lang.reflect.Array;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.management.JMException;
 import javax.management.MBeanAttributeInfo;
@@ -70,6 +74,8 @@ public class JmxDumpUtil
     public static final String PROTECTED_VALUE = "********";
 
     private static final String OS_NAME = "os.name";
+
+    private static final String INPUT_ARGUMENTS = "InputArguments";
 
     /**
      * Dumps a local or remote MBeanServer's entire object tree for support purposes. Nested arrays and CompositeData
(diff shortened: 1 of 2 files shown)


src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +170,70 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+    {
+        // Regex to select the characters following the words password=, token= or pwd=
+        String pattern = "(?i)((?<=(password)=)|(?<=(token)=)|(?<=(pwd)=)).++";
+        String replacement = JmxDumpUtil.PROTECTED_VALUE;
+        String output = input.replaceAll(pattern, replacement);
+
+        if(output.equals(input)) //If nothing has been replaced it's possible password= could be end of string
+        {
+            // pattern to see if the string ends with "password=", "token=" or "pwd="
+            Pattern passwordAtEnd = Pattern.compile("((password=)|(token=)|(pwd=))$");
```
eknizat: The pattern compilation is very expensive, it would be better to do this outside of this method. 
Also, can we configure this list of patterns or prop names somewhere? Maybe there is already a similar list defined?

ancutam: I agree with @eknizat, the pattern compiling can be done once, doesn't have to be done for every call. As for having a separate list, we do have a protected attributes list, but that applies to the mbean we manage. In this case this filters only the input arguments part of the default JVM mbean.
I think this covers what we want for now.
@skopf what do think?

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +170,70 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+    /**
+     * Removes any characters following the words:
+     * password
+     * pwd
+     * token
+     * and replaces them with JmxDumpUtil.PROTECTED_VALUE
+     * 
+     * Example: 
+     * Input:   -Ddb.password=alfresco
+     * Output:  -Ddb.password=JmxDumpUtil.PROTECTED_VALUE
```
ancutam: output is actually `-Ddb.password=*******`

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +170,70 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+     * 
+     * @param input
+     * @return password redacted string
+     */
+    public static String cleanPasswordFromInputArgument(String input)
+    {
+        // Regex to select the characters following the words password=, token= or pwd=
+        String pattern = "(?i)((?<=(password)=)|(?<=(token)=)|(?<=(pwd)=)).++";
```
ancutam: This can be extracted to a constant.

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
PunchedChimera: Was having issues with java doc, fixed now

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
PunchedChimera: Changed implementation of regex, it now has its own creation method: see lines 222 and lines 260-301

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
PunchedChimera: I have removed the used of `Pattern.compile` and modified the implementation of this method. The regex used now has its own creation method `createPasswordFindRegexString(String[] argEndings)` that can dynamically create the required regex, based on the argument provided.

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +169,136 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+            String arg = escapeRegexMetaChars(argEndings[i]);
+            part2 += "|" + arg + "="; //(.*(arg[0]=|arg[1]=arg[2]....|arg[n-1]=
+            part3 += "|" + arg + "="; //((?<=arg[0]=|arg[1]=|arg[2]...|arg[n-1]=
+        } 
+
+
+        part2 += "))"; //(.*(arg[0]=|arg[1]=|...|arg[n-1]=))
+        part3 += ").*+)"; //((?<=arg[0]=|arg[1]=|...|arg[n-1]=).*+)
+        String regex = part1 + part2 + part3;
+        return regex;
```
ancutam: Let's use something more elegant for string concatenation (maybe StringJoiner), make it easier to read/understand. 
I recommend this short article https://www.baeldung.com/java-strings-concatenation

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +169,136 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+     * </p>
+     * 
+     * @param input String
+     * @param redactedInputs String[]
+     * @return password redacted string if input matches a string in redactedInputs, an un-altered string will be returned if it does not match.
+     */
+    static String cleanPasswordFromInputArgument(String input, String[] redactedInputs)
+    {
+        //Selects the whole string, if one of the redactedInputs are present. It will have two capture groups, group 1 is the proceeding token ie. "password=" and group 2 will be all characters following the = (the vaule/password).
+        String regex = createPasswordFindRegexString(redactedInputs);
```
ancutam: This regex can be computed just once in `cleanPasswordsFromInputArguments`, since the `redactedInputs` list doesn't change between calls.

PunchedChimera: The regex is now computed just once, in `cleanPasswordsFromInputArguments`.

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
PunchedChimera: Creation of the string is now much simpler to read.

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +172,129 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+        {
+            IllegalArgumentException e = new IllegalArgumentException("Arguments are required");
+            throw e;
+        }
+
+        StringJoiner argJoiner = new StringJoiner("|");
+
+        for (String argEnding : argEndings) 
+        {
+            argJoiner.add(escapeRegexMetaChars(argEnding)+"=");
```
ancutam: You can add a prefix and a suffix using the joiner :)

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
```diff
@@ -164,9 +172,129 @@ private static void printMBeanInfo(MBeanServerConnection connection, ObjectName
...
+     * </p>
+     * 
+     * @param input String
+     * @param redactedInputs String[]
+     * @return password redacted string if input matches a string in redactedInputs, an un-altered string will be returned if it does not match.
+     */
+    static String cleanPasswordFromInputArgument(String input, Pattern redactedInputPattern)
+    {
+        //Replace the whole string with just capture group 1 to remove the desired value and concat the protected value.
+        String output = redactedInputPattern.matcher(input).replaceAll("$1"+PROTECTED_VALUE); //input.replaceAll(regex, "$1"+PROTECTED_VALUE);
```
ancutam: Please remove commented code.

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
PunchedChimera: Using suffix with the joiner would just add it to the end of the final joined string, which would output something like this: `"string1|string2|string3="` 
It needs to be `"string1=|string2=|string3="`

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
PunchedChimera: Done

src/main/java/org/alfresco/repo/management/JmxDumpUtil.java
ancutam: Ah ok, makes sense, sorry.

PR #536, opened 2019-07-30
[ SEARCH-1717 ] purgeOnInit flag defaults to false

agazzarini:
The PR contains a small change which sets the _purgeOnInit_ configuration property to false by default.


meenalbhave:
LGTM!

PR #537, opened 2019-07-30
REPO-4329 Configure Custom Renditions

alandavis:


alandavis:
Break done of changes:
- Removed rendition definition spring beans and created 0100-baseRenditions.json. Changes are picked up periodically by the running system.
- Refactored TransformServiceRegistryImpl to extract ConfigFileFinder to reads json from resources, files or directories.
- Refactored TransformServiceRegistryImpl to extract ConfigScheduler to periodically reads config data.
- Used ConfigFileFinder and ConfigScheduler in RenditionDefinition2Impl to read rendition defs.
- Removed the need for a current 'Data' parameter to be passed to register methods when adding transforms or renditions.
- Changes in test classes of AbstractRenditionIntegrationTests to force the config to be read once before each test, as tests were sometimes not getting the correct config due to scheduling.

montgolfiere:
@alandavis Just a quick early comment (I realise this is still marked as WIP). 

Can you please confirm that in addition to the new external json mechanism will these changes still be backwards compatible with existing mechanism (ie. spring bean xml-based config) ? In other words, is ATS-407 an optional or mandatory requirement for AIS 1.1 / ACS 6.2 (cc @DenisGabriela) ?

eg. https://github.com/Alfresco/alfresco-ai-renditions/blob/master/docker-compose/sample-custom-ai/custom-ai-service-context.xml

alandavis:
@montgolfiere You should still be able to use Spring beans to define your rendition definitions. They still self register when created and the registry is only using fields it creates itself to do that, so should work even if Spring has not finished setting all the properties on the registry itself.

That said, we already know that the AI amp is broken. It had to be commented out for the all amps test to run. ATS-407

tzclucian:
> @montgolfiere, @DenisGabriela You should still be able to use Spring beans to define your rendition definitions. They still self register when created and the registry is only using fields it creates itself to do that, so should work even if Spring has not finished setting all the properties on the registry itself.
> 
> That said, we already know that the AI amp is broken. It had to be commented out for the all amps test to run. ATS-407

@alandavis , I'm afraid this mechanism doesn't work anymore.

PR #568, opened 2019-08-20
REPO-4593: exclude duplicate library geronimo-jta.

anechifor:


anechifor:
There are different spec for geronimo-jta it's not a duplicate library as we thought.

PR #598, opened 2019-09-18
Cannot subclass TransformationOptionsConverter

Epurashu:
Change access modifier from private to protected for TransformationOptionsConverter#Setter interface and TransformationOptionsConverter#ifSet as TransformationOptionsConverter is required to be sub-classed by custom legacy Transformers.


Epurashu:
You are right, I should cherry pick the changes. Thank for the suggestions.

PR #599, opened 2019-09-18
REPO-4619 Allow additional parameters to be changed in content URLs

killerboot:


src/main/java/org/alfresco/repo/node/ContentPropertyRestrictionInterceptor.java
```diff
@@ -228,10 +239,7 @@ private boolean isContentNotNullOrEmpty(Serializable propValue)
         if (propValue instanceof ContentData)
         {
             String contentUrl = ((ContentData) propValue).getContentUrl();
-            if (contentUrl == null || contentUrl.isEmpty())
-            {
-                return false;
-            }
+            return contentUrl != null && !contentUrl.isEmpty();
```
killerboot: This is just a simplification of the expression.

montgolfiere:
👍

PR #643, opened 2019-10-18
REPO-4697/REPO-4701 Async transforms for Solr

alandavis:
- Async transform method added to RenditionService2
- Addition of a TransformDefinition so that a replyQueue and userData can be specified in a transform request. 
- Consume method in RenditionService2 has hook point to send response back to Solr if done Locally.
- ClientData enhanced to have extra fields for replyQueue and userData


src/main/java/org/alfresco/repo/rendition2/TransformEventProducer.java
```diff
@@ -0,0 +1,51 @@
...
+import org.alfresco.service.cmr.repository.NodeRef;
+
+import java.io.InputStream;
+
+/**
+ * Produce TransformReply events.
+ *
+ * @author aepure
+ */
+public interface TransformEventProducer
```
alandavis: See comment above about using Reply or Request rather than Event.

src/main/resources/alfresco/repository.properties
```diff
@@ -1307,6 +1307,9 @@ messaging.subsystem.autoStart=true
 # Raw events
 acs.repo.rendition.events.endpoint=jms:acs-repo-rendition-events?jmsMessageType=Text
```
alandavis: This one now probably needs a comment too.

PR #657, opened 2019-11-13
MNT-21009 :  Sandboxed  arbitrary Code Execution

anechifor:
Changed the initialization of secure scope.

### DIFF
--- a/src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
+++ b/src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
@@ -81,6 +81,9 @@ public class RhinoScriptProcessor extends BaseProcessor implements ScriptProcess
     /** Wrap Factory */
     private static final WrapFactory wrapFactory = new RhinoWrapFactory();
     
+    /** Sandbox Wrap Factory */
+    private static final SandboxWrapFactory sandboxFactory = new SandboxWrapFactory();
+    
     /** Base Value Converter */
     private final ValueConverter valueConverter = new ValueConverter();
     
@@ -453,14 +456,16 @@ public class RhinoScriptProcessor extends BaseProcessor implements ScriptProcess
         {
             // Create a thread-specific scope from one of the shared scopes.
             // See http://www.mozilla.org/rhino/scopes.html
-            cx.setWrapFactory(wrapFactory);
+            
             Scriptable scope;
             if (this.shareSealedScopes)
             {
                 Scriptable sharedScope = secure ? this.nonSecureScope : this.secureScope;
+                cx.setWrapFactory(secure ? wrapFactory : sandboxFactory);
                 scope = cx.newObject(sharedScope);
                 scope.setPrototype(sharedScope);
                 scope.setParentScope(null);
+
             }
             else
             {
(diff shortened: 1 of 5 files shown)


src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
```diff
@@ -643,27 +643,27 @@ public void afterPropertiesSet() throws Exception
...
+    protected Scriptable initScope(Context cx, boolean secure, boolean sealed)
+    {
+        Scriptable scope;
+        if (secure)
+        {
+            // Initialise the non-secure scope
+            // allow access to all libraries and objects, including the importer
+            // @see http://www.mozilla.org/rhino/ScriptingJava.html
+            scope = new ImporterTopLevel(cx, sealed);
+        }
```
skopf: What is the meaning of the "secure" parameter here? Shouldn't it be the other way round?

anechifor: I didn't change nothing on that part. The meaning of secure in this case is that we consider code to be secure for exemple using javascript as string or javascript files from classPath. I didn't change this part. If you take a look you will se that I change (insecure) using Rhino  initSafeStandardObjects this is used for javascript inside nodeRef. Please see: https://mozilla.github.io/rhino/javadoc/org/mozilla/javascript/Context.html#initSafeStandardObjects-org.mozilla.javascript.ScriptableObject.

anechifor: The main idea is that Rhino introduce from 1.2.7 the concept of sandbox as you mentioned in the jira and I used. I didn't add parameters because I repaired the delete("java") with the call of initSafe.....

src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
```diff
@@ -593,6 +598,24 @@ public Scriptable wrapAsJavaObject(Context cx, Scriptable scope, Object javaObje
...
             return super.wrapAsJavaObject(cx, scope, javaObject, staticType);
         }
+    }
+    
+    
+    /**
+     * A {@link WrapFactory} that ensures {@link org.mozilla.javascript.NativeJavaObject} instances are of the
+     * {@link SandboxNativeJavaObject} variety.
+     */
+    private static class SandboxWrapFactory extends WrapFactory
```
skopf: 1. Why do we have a special RhinoWrapFactory?
2. Shouldn't then our SandboxWrapFactory extend this RhinoWrapFactory

src/main/java/org/alfresco/repo/jscript/RhinoScriptProcessor.java
```diff
@@ -453,14 +456,16 @@ private Object executeScriptImpl(Script script, Map<String, Object> model, boole
...
             // Create a thread-specific scope from one of the shared scopes.
             // See http://www.mozilla.org/rhino/scopes.html
-            cx.setWrapFactory(wrapFactory);
+            
             Scriptable scope;
             if (this.shareSealedScopes)
             {
-                Scriptable sharedScope = secure ? this.nonSecureScope : this.secureScope;
+                Scriptable sharedScope = secure ? this.nonSecureScope : this.secureScope;
+                cx.setWrapFactory(secure ? wrapFactory : sandboxFactory);
```
skopf: 1. Why has this line been moved from outside of the if(this.shareSealedScops) block into this if block?
2. If this.shareSealedScopes is false, then cx.setWrapFactory() does not get invoked

PR #658, opened 2019-11-14
MNT-18497 - only add cm:properties to the json if it's a link, skip other properties

astrachan:


KikoUK:
Changes in the PR do not fix the issue. This is due to the fact that the additional properties and aspects are a result of the creation of a new object within the linkedNode object in this part of the code:
if (nodeInfo.isLink())
        {
            NodeRef targetNodeRef = nodeInfo.getLinkNodeRef();
            if (targetNodeRef != null)
            {
                rootJSONObject.put("linkedNode", -->toJSONObject<--(targetNodeRef, useShortQNames));
            }
        }
So the "problematic" props an aspects are never reached with the code from the PR as the creation of the targetRef object happens earlier in the call stack and when it reaches the suggested code .isLink() returns false due to the current object being the original file which the link points to.

PR #697, opened 2019-12-05
REPO-4791 Use Local transforms where Legacy transforms are called.

alandavis:
- Re factored TransformerDebug so the Legacy transform code has been extracted into
  LegacyTransformerDebug and code used by the Admin console's Test Transform that is
  not being deprecated has been extracted into AdminUiTransformerDebug
- Removed a few tests for a method that no longer supports specifying a transform name.
- All other changes are to handle the fact that the transformerDebug bean is now a
  LegacyTransformerDebug class and methods have been moved to sub classes.


alandavis:
Although TransformerDebug appears to have had major changes in this PR, these are limited to just splitting the class into three. In doing so some of the methods are no longer private but now protected. The idea is that at some point we will be able to simply remove LegacyTransformerDebug which has references to Legacy transformer classes.

PR #710, opened 2019-12-06
REPO-4816 - Remove library edu.ucar:netcdf from alfresco-repository p…

antoniojfelix:
…roject
Regarding the issue REPO-4695, this is a library that can be removed according with the analysis done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -578,12 +578,6 @@
             <version>2.0.7</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


Epurashu:
For investigation regarding edu.ucar:netcdf was raised an issue quite some time ago -> https://issues.alfresco.com/jira/browse/REPO-3967 according to the current investigation, this library can be removed.

PR #724, opened 2019-12-13
MNT-20992: Don't cache empty property maps

eknizat:
Property values are used as keys in caches but maps don't have a good enough equals & hashcode implementations to be used as keys. This caused wrong types of values being recorded in the database. This commit excludes empty maps from cacheing in property value dao operations.


src/main/java/org/alfresco/repo/domain/propval/AbstractPropertyValueDAOImpl.java
```diff
@@ -784,6 +784,12 @@ public Serializable getValueKey(Serializable value)
                 // It will be Serialized, so no search key
                 return null;
             }
+            else if (value instanceof Map<?, ?> && ((Map) value).isEmpty())
```
nithinnambiar: Type argument is not required, Map  instead of Map<?, ?> would be sufficient.

eknizat: You are correct. The wildcard type parameter here doesn't buy us anything, but I wen't with it to keep to the existing style already present in this and surrounding classes.

ancutam:
Looks good. Thank you @eknizat for the thorough investigation.

PR #795, opened 2020-01-31
SEARCH-2074: Add new queries tests to ensure these ANSI SQL queries w…

aborroy:
…ork with different database engines.


aborroy:
Check returned values from queries instead of a simple not null validation.

PR #808, opened 2020-02-12
MNT-21083: Alternative clean up script

cristianturlica:
- initial alternative clean up script


src/test/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsExecutorTest.java
```diff
@@ -0,0 +1,183 @@
...
+import org.mockito.Mockito;
+import org.springframework.context.ApplicationContext;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Integration tests for the {@link DeleteNotExistsExecutor} class.
+ * 
+ * @author Cristian Turlica
+ */
+@Category(LuceneTests.class)
```
alandavis: LuceneTests?

src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsExecutor.java
```diff
@@ -0,0 +1,555 @@
...
+                {
+                    batchSizeString = globalProperties.getProperty(PROPERTY_DEFAULT_BATCH_SIZE);
+                }
+
+                batchSize = batchSizeString == null ? 10000 : Integer.parseInt(batchSizeString);
+
+
+                // Read the batch size from the named property
+                String deleteBatchSizeString = globalProperties.getProperty(PROPERTY_DEFAULT_DELETE_BATCH_SIZE);
+                deleteBatchSize = deleteBatchSizeString == null ? 10000 : Integer.parseInt(deleteBatchSizeString);
```
alandavis: 10,000 sounds a lot. I would have guessed a smaller number would work well.

src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsExecutor.java
```diff
@@ -0,0 +1,555 @@
...
+        if (deleteIds.isEmpty()) {
+            return;
+        }
+
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("Prepare to delete " + deleteIds.size() + " items from table " + tableName + ".");
+        }
+
+        StringBuilder stmtBuilder = new StringBuilder("DELETE FROM " + tableName + " WHERE " + idColumnName + " IN ");
```
alandavis: It is possible to do this with a prepared statement too. Generally half the processing time is preparing, so is worth doing if it is run many times. Very long statements may also be an issue with some DBs. Where you don't have a full set of IDs for your prepared statement, you would just repeat one of the IDs many times when you bind variables to the prepared statement.

src/test/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsExecutorTest.java
cristianturlica: Removed. Thanks.

src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsExecutor.java
cristianturlica: Added default of 500 but it can be configured to something else via properties.

src/main/java/org/alfresco/repo/domain/schema/script/DeleteNotExistsExecutor.java
cristianturlica: Changes done

PR #815, opened 2020-02-14
REPO-4710 Remove Legacy Transforms

alandavis:
The Local transformers should support more of the legacy transforms.

- Addition of 'wildcard' supported transformations in pipelines, based on the transformers and intermediates types being used. Greatly increases the number of transforms supported by local transforms so it is similar to legacy.
- Renamed test transform option names to match those used in the real T-Engines
- Consume aiToImageViaPng into pdfToImageViaPng as that was the pattern in legacy and will make building pipelines simpler.
- Modified legacy transformers complex.Rfc822ToPdf and complex.ArchiveToPdf so that they only go to pdf. The other 140 combinations were added in error a couple of years ago.
- Enhance TransformerDebug transformations by extension, to show multiple Local transformers if they have different priorities or source size limits.


src/main/java/org/alfresco/repo/content/transform/LegacyTransformerDebug.java
```diff
@@ -496,11 +497,11 @@ public int compare(ContentTransformer transformer1, ContentTransformer transform
                                 ? false
                                 : remoteTransformServiceRegistry.isSupported(sourceMimetype,
                                 -1, targetMimetype, Collections.emptyMap(), null);
-                        LocalTransform localTransform = localTransformServiceRegistryImpl == null
-                                ? null
-                                : localTransformServiceRegistryImpl.getLocalTransform(sourceMimetype,
-                                -1, targetMimetype, Collections.emptyMap(), null);
-                        if (localTransform != null || supportedByTransformService || size >= 1)
+                        List<SupportedTransform> localTransformers = localTransformServiceRegistryImpl == null
```
alandavis: The code changes in this class are very similar to those in the super class. The two classes repeat much of the same code rather than having common code in the super class as the plan is to simply remove this class shortly. As a result the super class remains simpler to understand.

src/main/java/org/alfresco/repo/content/transform/LocalPassThroughTransform.java
```diff
@@ -76,10 +76,10 @@ public static Transformer getConfig(List<String> mimetypes)
         Set<SupportedSourceAndTarget> supportedSourceAndTargetList = new HashSet();
         for (String mimetype: mimetypes)
         {
-            supportedSourceAndTargetList.add(new SupportedSourceAndTarget(mimetype, mimetype, -1));
+            supportedSourceAndTargetList.add(new SupportedSourceAndTarget(mimetype, mimetype, -1, 20));
```
alandavis: Bug Fix: Legacy pass through had a priority of 20. Without this we would sometimes use more complex/costly transforms.

src/main/resources/alfresco/subsystems/Transformers/default/transformers.properties
```diff
@@ -61,10 +61,8 @@ content.transformer.complex.Any.Image.priority=400
 
 content.transformer.complex.Text.Image.pipeline=*|pdf|complex.PDF.Image
 content.transformer.complex.Text.Image.priority=350
-content.transformer.complex.Text.Image.extensions.csv.*.supported=true
```
alandavis: Source types of csv and txt were never used in Legacy. When looking at what should be supported in Local this complicated the situation.

src/main/resources/alfresco/subsystems/Transformers/default/transformers.properties
```diff
@@ -102,9 +100,11 @@ content.transformer.TikaAuto.extensions.pdf.txt.maxSourceSizeKBytes=25600
 
 # EML to PDF
 content.transformer.complex.Rfc822ToPdf.pipeline=RFC822|txt|*
+content.transformer.complex.Rfc822ToPdf.extensions.*.pdf.supported=true
```
alandavis: Bug Fix: Legacy transformer should only have been going to pdf.

src/main/resources/alfresco/transforms/0100-basePipelines.json
```diff
@@ -7,27 +7,6 @@
         {"transformerName": "imagemagick"}
       ],
       "supportedSourceAndTargetList": [
-        {"sourceMediaType": "application/pdf",                    "targetMediaType": "image/gif" },
```
alandavis: These two transformers don't need to be separate, and makes using them in other pipelines harder. i.e. you need extra containing pipelines.

src/main/resources/alfresco/transforms/0100-basePipelines.json
```diff
@@ -154,7 +111,7 @@
         {"transformerName": "libreoffice"}
       ],
       "supportedSourceAndTargetList": [
-        {"sourceMediaType": "text/html",                    "targetMediaType": "application/pdf" }
+        {"sourceMediaType": "text/html",                    "priority": "150", "targetMediaType": "application/pdf" }
```
alandavis: Think this change needs to be removed. Will need to check tests with a build.

src/test/java/org/alfresco/repo/rendition2/LocalTransformServiceRegistryIntegrationTest.java
```diff
@@ -92,30 +95,34 @@ protected boolean isEnabled()
...
     public void testIsSupported()
     {
+        // Need to make sure we don't fall back to a transformer without limits.
+        List<SupportedTransform> transformers = localTransformServiceRegistry.findTransformers(MIMETYPE_IWORK_PAGES, targetMimetype, options, RENDITION_NAME);
+        assertEquals(1, transformers.size());
+
         // +ve
         // No props
-        Assert.assertTrue(transformServiceRegistry.isSupported(MIMETYPE_OPENXML_WORDPROCESSING, 1234, MIMETYPE_PDF, options, RENDITION_NAME));
+        Assert.assertTrue(transformServiceRegistry.isSupported(MIMETYPE_IWORK_PAGES, 1234, targetMimetype, options, RENDITION_NAME));
```
alandavis: The wildcard changes introduced another transformer from docx to pdf which did not have a size limit. As a result this test failed. Ended up having to create a 'dummy' test pipeline (only in the test context) that could be used, hence the need to change mimetypes.

src/test/java/org/alfresco/transform/client/registry/LocalTransformServiceRegistryConfigTest.java
```diff
@@ -191,8 +191,20 @@ protected String getTransformServiceConfigPipeline()
...
+// pdfrendere
+//        {"sourceMediaType": "application/pdf", "targetMediaType": "image/png" }
+// libreoffice
+//         {"sourceMediaType": "application/vnd.ms-outlook", "targetMediaType": "application/pdf"}
+// officeToImageViaPdf
+//        {"sourceMediaType": "application/msword",  "targetMediaType": "image/gif" },
+//        {"sourceMediaType": "application/msword",  "targetMediaType": "image/jpeg"},
+//        {"sourceMediaType": "application/msword",  "targetMediaType": "image/png" },
+//        {"sourceMediaType": "application/msword",  "targetMediaType": "image/tiff"}
+        return 3 + 1 + 1 + 4;   // 18
```
alandavis: comment needs correction 9 not 18.

src/test/java/org/alfresco/transform/client/registry/TransformServiceRegistryConfigTest.java
```diff
@@ -76,6 +76,13 @@ public void setUp() throws Exception
     {
         super.setUp();
         LogManager.getLogger(TransformServiceRegistryConfigTest.class).setLevel(Level.DEBUG);
+        System.setProperty("localTransform.appleIWorks.url", "http://imagemagick:8090/");
```
alandavis: Can remove. Will run a build without this code.

src/test/resources/alfresco/local-transform-service-config-pipeline-test.json
```diff
@@ -1,6 +1,6 @@
 {
   "transformOptions": {
-    "imagemagickOptions": [
```
alandavis: Switch to using the production T-Engine option names

src/test/resources/alfresco/transforms/0999-testPipeline.json
```diff
@@ -0,0 +1,16 @@
+{
```
alandavis: Test context dummy pipeline that allows us to test the availability of transformers with different source sizes. No longer possible with the standard OOTB config.

PR #825, opened 2020-02-22
REPO-4710 Prep for Remove Legacy Transformers 

alandavis:
Bug Fix: Test pick up config resources on both main and test classpaths


alandavis:
Tests were green, so can close

PR #849, opened 2020-03-04
MNT-20822 Delete CMISTest.testQueryNodesWithCorruptedIndexes.

tpage-alfresco:
This test was written for Lucene systems, and no longer applies to Solr systems.

This change also deletes the unused logger and factory classes.


eliaporciani:
LGTM

PR #863, opened 2020-03-06
REPO-4963 - Add direct dependencies of io.netty libraries to sync the…

antoniojfelix:
… same version


ancutam:
reverted in #891

PR #889, opened 2020-03-18
REPO-4963 - Remove provided tag and add netty lib exclusion to avoid …

antoniojfelix:
…different lib versions


ancutam:
reverted in #891

Issue #922, opened 2020-04-06
The display name of the scheduled job is incorrect.


silverskyvicto:
Hi.

When accessing the URL of `<Base URL>/alfresco/service/enterprise/admin/admin-scheduledjobs` and viewing the list of scheduled jobs, the display name of the job looks like a Java memory address.

The following image is attached as an example.

![scheduled_job_list](https://user-images.githubusercontent.com/3609003/78533460-f3eff800-7823-11ea-9667-588e884043c6.png)

PR #923, opened 2020-04-06
[#922] add scheduled job name.

silverskyvicto:
HI.

There was a problem that the display name of some scheduled jobs was wrong, and I tried to fix it.

I am just modifying the `xml` file.


silverskyvicto:
Hi.

How about this issue?

PR #930, opened 2020-04-08
REPO-5016 Add Node events to the content repository

jamalkm:


src/test/java/org/alfresco/repo/event2/EventGeneratorTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.alfresco.util.BaseSpringTest;
+import org.alfresco.util.GUID;
+import org.alfresco.util.PropertyMap;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+
+public class EventGeneratorTest extends BaseSpringTest
```
aftene, line 47: I think is not necessary to commit this test class.

src/test/java/org/alfresco/repo/event2/DeleteRepoEventIT.java
```diff
@@ -0,0 +1,74 @@
...
+        final RepoEvent<NodeResource> resultRepoEvent = getRepoEvent(2);
+
+        assertEquals("Repo event type:", EventType.NODE_DELETED.getType(), resultRepoEvent.getType());
+        assertEquals(createdResource.getId(), getNodeResource(resultRepoEvent).getId());
+
+        // There should be no resourceBefore
+        EventData<NodeResource> eventData = getEventData(resultRepoEvent);
+        assertNull("There should be no 'resourceBefore' object for the Deleted event type.",
+                   eventData.getResourceBefore());
+    }
```
alandavis, line 57: Iulian: Need to include a sample of asserts that there were properties, aspects... before the delete.

src/test/java/org/alfresco/repo/event2/DeleteRepoEventIT.java
```diff
@@ -0,0 +1,74 @@
...
+        createNode(ContentModel.TYPE_CONTENT, parent);
+
+        // 4 Created Events
+        checkNumOfEvents(4);
+
+        deleteNode(grandParent);
+        // 4 Deleted events + 4 created events
+        checkNumOfEvents(8);
+    }
+}
```
alandavis, line 74: Iulian: Can we please have a test where a node is created and deleted in the same transaction, so does not generate an event. It is fine if there is another event in the transaction.

alandavis, line 74: Iulian: Can we have a test where there are multiple events from a single transaction, because there are multiple nodes. Happy if it is combined with the temporary node case. Another class?

src/main/java/org/alfresco/util/PropertyMap.java
```diff
@@ -23,197 +23,200 @@
...
-        return result;
-    }
-}
+package org.alfresco.util;
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
```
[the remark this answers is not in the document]
jamalkm: It seems the file has been changed from CRLF to LF.

src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
```diff
@@ -0,0 +1,277 @@
...
+                {
+                    NodeRef pNodeRef = personService.getPersonOrNull(userName);
+                    if (pNodeRef != null)
+                    {
+                        pInfo = personService.getPerson(pNodeRef);
+                    }
+                }
+                catch (NoSuchPersonException | AccessDeniedException ex)
+                {
+                    // ignore
```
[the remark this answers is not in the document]
jamalkm: Will add the test in the next iteration.

PR #932, opened 2020-04-09
REPO-4859 : HTTP_UNAUTHORIZED instead of HTTP_FORBIDDEN for some CMIS apis

tzclucian:
- updated code so the Context set includes the USERNAME field.
    - the missing of that field on the context was causing the CMIS lib to trigger a 401

## Long story short
The 401 is triggered at this line, because the username is empty.

Screenshot 2020-04-09 at 13 45 38

The username is empty because it is never set. The created context is an instance of `BrowserCallContextImpl`. No piece of code/logic that is being executed is putting the `USERNAME` in the context.

The context is created as you can see below in `CmisBrowserBindingServlet`.

Screenshot 2020-04-09 at 13 47 15

Immediately after the context is created the request is being dispatched.
After the dispatch, we have access through the Alfresco code to the context.

First in `AlfrescoCmisServiceFactory` which opens the service as per below.
Screenshot 2020-04-09 at 13 50 07

As the `AlfrescoCmisServiceFactory` is no place for modifying the context, I've added the fix logic in the `AlfrescoCmisServiceImpl` class.

It is in my opinion, the first moment when we can add this.
Before the context is created there is no way of 'forcing' the construction logic for the`BrowserCallContextImpl` to include the `USERNAME` field.

Therefor, this is why I think this is where the addition of the username should happen.

### DIFF
--- a/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
+++ b/src/main/java/org/alfresco/opencmis/AlfrescoCmisServiceImpl.java
@@ -44,8 +44,6 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import javax.servlet.http.HttpServletRequest;
 
-import net.sf.acegisecurity.Authentication;
-
 import org.alfresco.model.ContentModel;
 import org.alfresco.opencmis.dictionary.CMISNodeInfo;
 import org.alfresco.opencmis.dictionary.CMISObjectVariant;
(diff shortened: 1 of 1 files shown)


tzclucian:
Kind reminder.

ancutam:
Looks good to me.
Any chance to add a TAS test for this, we have a suite running CMIS tests against Alfresco and AIMS (in acs-packaging)?

PR #919, opened 2020-04-03
MNT-20764: Searches fail for users who are members of groups where the authorityName contains double quotes

amohammedalfresco:
- Guard AuthorityServiceImpl against an array of illegal characters while creating group and added a test for it

PR #933, opened 2020-04-09
MNT-20764: Searches fail for users who are members of groups where the authorityName contains double quotes

amohammedalfresco:
- Throw exceptions when a group name contain invalid character (`'/', '\\', '\n', '\r', '"'`)


amohammedalfresco:
Closed #919 due to EOL issues

src/test/java/org/alfresco/repo/security/authority/AuthorityServiceTest.java
```diff
@@ -252,6 +252,20 @@ public void testMNT10533() throws Exception
         }
     }
 
+    public void testCreateAuthorityWithIllegalCharacters() throws Exception
+    {
+        char[] illegalCharacters = {'/', '\\', '\n', '\r', '"'};
+        for (char illegalCharacter : illegalCharacters)
+        {
+            String GROUP_NAME = "testGroupNameWith" + illegalCharacter;
+            try {
```
killerboot, line 261: Please check your code style. Braces are in wrong places.

src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
```diff
@@ -1046,6 +1046,17 @@ else if (type.isPrefixed())
             final String authorityName = prefix + shortName;
             if (shortName.startsWith(prefix))
             {
+                if (type == AuthorityType.GROUP)
```
killerboot, line 1049: Are these characters allowed for users?

src/test/java/org/alfresco/repo/security/authority/AuthorityServiceTest.java
```diff
@@ -351,9 +365,14 @@ public void testZones()
     @Category(RedundantTests.class)
     public void test_ETWOTWO_400()
```
killerboot, line 366: I think this test needs to be deleted as it is essentially the same as testCreateAuthorityWithIllegalCharacters now.

src/test/java/org/alfresco/repo/security/authority/AuthorityServiceTest.java
amohammedalfresco: Thanks for the spotting it, just linted it.

src/test/java/org/alfresco/repo/security/authority/AuthorityServiceTest.java
amohammedalfresco: Removed the test.

src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
amohammedalfresco: I thought of initially adding it for the `AuthorityType.USER` but the [createAuthority](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/security/authority/AuthorityServiceImpl.java#L645) method calls another method [checkTypeIsMutable](https://github.com/Alfresco/alfresco-repository/blob/master/src/main/java/org/alfresco/repo/security/authority/AuthorityServiceImpl.java#L527) and it has explicit check for either `AuthorityType.GROUP` or `AuthorityType.ROLE` which makes me to believe this `createAuthority` method may not be creating `AuthorityType.USER`.  Hence, I only guarded the `checkTypeIsMutable` method with `AuthorityType.Group`.

killerboot: I think it is required to extend the scope of this change to make sure that Users and Groups can not be created with wrong characters using Java APIs.

src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
```diff
@@ -1656,6 +1657,28 @@ public void onUpdateProperties(NodeRef nodeRef, Map<QName, Serializable> before,
         }
     }
 
+    public void onCreateNode(ChildAssociationRef childAssocRef)
+    {
+        // Restrict creation of group name that contain invalid characters.
+        NodeRef groupRef = childAssocRef.getChildRef();
```
killerboot, line 1663: Please change the variables to `authorityRef` and `authorityName`

src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
```diff
@@ -1656,6 +1657,28 @@ public void onUpdateProperties(NodeRef nodeRef, Map<QName, Serializable> before,
...
     }
 
+    public void onCreateNode(ChildAssociationRef childAssocRef)
+    {
+        // Restrict creation of group name that contain invalid characters.
+        NodeRef groupRef = childAssocRef.getChildRef();
+
+        String groupName = (String) this.nodeService.getProperty(groupRef, ContentModel.PROP_AUTHORITY_NAME);
+
+        if (getAuthorityContainer().equals(childAssocRef.getParentRef()))
```
killerboot, line 1667: Is this check required?

src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
amohammedalfresco: Done.

src/main/java/org/alfresco/repo/security/authority/AuthorityDAOImpl.java
amohammedalfresco: Removed the not needed check.

PR #939, opened 2020-04-15
MNT-21417: LDAP Sync not working after making changes via admin UI

ancutam:
Use `AlfrescoSchedulerAccessorBean` instead of `SchedulerAccessorBean`, to ensure the ldap sync job is unscheduled when the 'Synchronization' subsystem is restarted.
Add a bean name for the sync trigger job details - `ldapPeopleJobDetail`, this will appear as job name in Admin console > Scheduled Jobs.


ancutam:
TAS tests added here https://github.com/Alfresco/acs-packaging/pull/1135

PR #940, opened 2020-04-15
MNT-19887: Non-responsive SOLR address solr breaks admin console pages

CezarLeahu:
- add timeout on SOLR Admin HTTP call
- wire *solrConnectTimeout* (5 sec) in the solrAdminHttpClient beans
- add unit tests for the timeout logic (SolrAdminHTTPClientTest)
- refactor SolrChildApplicationContextFactory
- add *recent_timeout_exception_cache* for SOLR Timeouts
- add unit tests for the recent_timeout_exception_cache (SolrChildApplicationContextFactoryTest)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ jobs:
     - stage: test
       name: "AllUnitTestsSuite"
       script: mvn test -B -Dtest=AllUnitTestsSuite
+
     - name: "WhiteSource scan"
       # only on support branches or master and if it is not a PR
       if: fork = false AND (branch = master OR branch =~ /support\/SP\/.*/) AND type != pull_request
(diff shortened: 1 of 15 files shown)


CezarLeahu:
@killerboot Thanks for the review. I missed the httpClientFactory configuration and was under the impression I can't alter the httpClient behavior. I'll redo the fix.

As for how the _Search Service_ admin page would look like if solr is unreachable: pretty much the same as when it is reachable, but with missing information and no explicit error message. All the fields (14 in total) from the **Store Tracking Status** sections would have "_Information is currently unavailable_" as values. There's been no changes on the values returned in case of errors/exceptions; it's just that they're now returned quicker.

CezarLeahu:
I'm closing the PR while I'm addressing the comments and re-implementing the fix - which will look quite different when done.

PR #942, opened 2020-04-16
MNT-21328 : Delete Share site can be recreated with same name while deleted site is in trashcan

DenisGabriela:
- check authorities for given shortName with ignore case filter
   - add tests

### DIFF
--- a/src/main/java/org/alfresco/repo/security/authority/script/ScriptAuthorityService.java
+++ b/src/main/java/org/alfresco/repo/security/authority/script/ScriptAuthorityService.java
@@ -385,7 +385,7 @@ public class ScriptAuthorityService extends BaseScopableProcessorExtension
         PagingResults<String> authorities = authorityService
             .getAuthorities(AuthorityType.GROUP, null, displayNameFilter, false, false,
                 new PagingRequest(0, 1, null));
-        return authorities != null && !authorities.getPage().isEmpty() ? true : false;
+        return authorities != null && !authorities.getPage().isEmpty();
     }
 
 	/**
--- a/src/main/java/org/alfresco/repo/security/authority/script/ScriptAuthorityService.java
+++ b/src/main/java/org/alfresco/repo/security/authority/script/ScriptAuthorityService.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.alfresco.model.ContentModel;
+import org.alfresco.query.PagingRequest;
 import org.alfresco.query.PagingResults;
 import org.alfresco.repo.jscript.BaseScopableProcessorExtension;
 import org.alfresco.repo.security.authority.AuthorityInfo;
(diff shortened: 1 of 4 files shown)


src/main/java/org/alfresco/repo/site/SiteServiceImpl.java
```diff
@@ -1493,7 +1491,29 @@ public NodeRef doWork() throws Exception
...
+
+            List<ChildAssociationRef> childAssociationRefs = directNodeService
+                .getChildAssocs(storeArchiveNode, ContentModel.ASSOC_CHILDREN,
+                    NodeArchiveService.QNAME_ARCHIVED_ITEM);
+
+            Optional<NodeRef> archivedShortNameSite = childAssociationRefs.stream()
+                .map(childAssociationRef -> childAssociationRef.getChildRef())
+                .filter(site -> ((String) directNodeService.getProperty(site, ContentModel.PROP_NAME))
+                    .equalsIgnoreCase(shortName))
+                .findFirst();
```
cristianturlica, line 1511: We should try to seek a different approach to find if the node is already archived. In the current approach, if we have a large amount of items in the trashcan then it will retrieve and iterate on all...

src/main/java/org/alfresco/repo/security/authority/script/ScriptAuthorityService.java
```diff
@@ -372,7 +373,21 @@ public ScriptGroup getGroupForFullAuthorityName(String fullAuthorityName)
...
+     * @return true if there is at least an authority starting with given displayNameFilter, false otherwise
+     */
+	public boolean authoritiesExist(String displayNameFilter)
+    {
+        PagingResults<String> authorities = authorityService
+            .getAuthorities(AuthorityType.GROUP, null, displayNameFilter, false, false,
+                new PagingRequest(0, 1, null));
+        return authorities != null && !authorities.getPage().isEmpty();
+    }
+
```
cristianturlica, line 390: Is this needed? we have the same method as private in SiteServiceImpl...

DenisGabriela: Thanks for review. This method will be used by the `sites.post` webscript in alfresco-remote-api -> https://github.com/Alfresco/alfresco-remote-api/pull/609/files#diff-43ccf775091f54ebf7aad38dfe4631f1R26 (next PR to be reviewed - https://github.com/Alfresco/alfresco-remote-api/pull/609). The webscript is not using `SiteServiceImpl` directly, but through `ScriptSiteService`. It is also using `ScriptAuthorityService` for authority specific calls. Thus if I want to use authoritiesExist(displayNameFilter) method in `sites.post` it needs to be defined in one of the script services mentioned above and `ScriptAuthorityService` is more appropriate for that. Let me know if this makes sense or there is a better solution.

DenisGabriela:
Details in Jira: https://issues.alfresco.com/jira/browse/MNT-21328?focusedCommentId=767377&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-767377

PR #943, opened 2020-04-19
REPO-5079,ATS-707: Use all-in-one T-Engine and fix tests

alandavis:


alandavis:
WIP: Going to switch back to a separate .url for each of the legacy T-Engines so the chart in acs-deployment can point legacy transforms to separate T-Engines. Otherwise we would need a AIO there in addition for testing.

montgolfiere:
fyi @eknizat

alandavis:
Was approved by Erik

PR #968, opened 2020-04-29
REPO-5101-Update Repo Event Integration Tests

aftene:
- add create-delete node in the same transaction
- add create-update node in the same transaction
- add create multiple nodes in the same transaction
- validate properties before deleting a node

### DIFF
--- a/src/test/java/org/alfresco/repo/event2/AbstractContextAwareRepoEvent.java
+++ b/src/test/java/org/alfresco/repo/event2/AbstractContextAwareRepoEvent.java
@@ -265,7 +265,7 @@ public abstract class AbstractContextAwareRepoEvent extends BaseSpringTest
      */
     protected void waitUntilNumOfEvents(int numOfEvents)
     {
-        await().atMost(1000, SECONDS).until(() -> EVENT_CONTAINER.getEvents().size() == numOfEvents);
+        await().atMost(5, SECONDS).until(() -> EVENT_CONTAINER.getEvents().size() == numOfEvents);
     }
 
     protected void checkNumOfEvents(int expected)
--- a/src/test/java/org/alfresco/repo/event2/AbstractContextAwareRepoEvent.java
+++ b/src/test/java/org/alfresco/repo/event2/AbstractContextAwareRepoEvent.java
@@ -71,7 +71,8 @@ import org.springframework.beans.factory.annotation.Qualifier;
 
 public abstract class AbstractContextAwareRepoEvent extends BaseSpringTest
 {
-    private static final String     TEST_NAMESPACE = "http://www.alfresco.org/test/ContextAwareRepoEvent";
+    protected static final String     TEST_NAMESPACE = "http://www.alfresco" +
+                                                ".org/test/ContextAwareRepoEvent";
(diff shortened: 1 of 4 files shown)


src/test/java/org/alfresco/repo/event2/UpdateRepoEventIT.java
```diff
@@ -281,4 +282,23 @@ public void removeAspectFromContentTest()
         assertNotNull(resourceBefore.getAspectNames());
         assertEquals(aspectsBeforeRemove, resourceBefore.getAspectNames());
     }
+
+    @Test
+    public void testCteateAndUpdateInTheSameTransaction()
```
jamalkm, line 287: typo: Ctreate -> Create

src/test/java/org/alfresco/repo/event2/DeleteRepoEventIT.java
```diff
@@ -44,6 +46,14 @@ public void deleteContent()
         NodeRef nodeRef = createNode(ContentModel.TYPE_CONTENT);
```
jamalkm, line 46: Please add "cm:title" property when creating the node so we get a non-empty properties map.

src/test/java/org/alfresco/repo/event2/AbstractContextAwareRepoEvent.java
```diff
@@ -264,7 +265,7 @@ protected NodeResource getNodeResourceBefore(RepoEvent<NodeResource> repoEvent)
      */
     protected void waitUntilNumOfEvents(int numOfEvents)
     {
-        await().atMost(5, SECONDS).until(() -> EVENT_CONTAINER.getEvents().size() == numOfEvents);
+        await().atMost(1000, SECONDS).until(() -> EVENT_CONTAINER.getEvents().size() == numOfEvents);
```
jamalkm, line 268: Do we really need to make it 1000s? I think 5s or max 10s is more than enough. 
Also, the method JavaDoc needs updating if the wait time is not 5 second.

aftene: I've forgot to change it back after debug, sorry.

src/test/java/org/alfresco/repo/event2/CreateRepoEventIT.java
```diff
@@ -146,4 +147,33 @@ public void testFileNodeType()
...
+        RepoEventContainer repoEventsContainer = getRepoEventsContainer();
+        final String eventGroupId1 =
+            getEventData(repoEventsContainer.getEvent(1)).getEventGroupId();
+        final String eventGroupId2 =
+            getEventData(repoEventsContainer.getEvent(2)).getEventGroupId();
+        final String eventGroupId3 =
+            getEventData(repoEventsContainer.getEvent(3)).getEventGroupId();
+
+        //All events in the transaction should have the same eventGroupId
+        assertTrue(eventGroupId1.equals(eventGroupId2) || eventGroupId2.equals(eventGroupId3));
```
jamalkm, line 177: I forgot to add a comment here earlier, but the logic is wrong, it should be an `&&` condition

PR #974, opened 2020-04-30
REPO-4859 : HTTP_UNAUTHORIZED instead of HTTP_FORBIDDEN for some CMIS…

tzclucian:
- moved the fix to a more suitable place


src/main/java/org/alfresco/opencmis/PublicApiCallContextHandler.java
```diff
@@ -46,8 +48,12 @@
         {
             map.putAll(basicAuthMap);
         }
-        
-        map.put("isPublicApi", "true");
-        return map;
-	}
-}
+
```
cristianturlica, line 51: Please add a comment to detail what happens in this case.

PR #990, opened 2020-05-14
REPO-5121 Node Type Changed Event

SaraAspery:


src/main/java/org/alfresco/repo/event2/EventConsolidator.java
```diff
@@ -265,6 +274,11 @@ private NodeResource buildNodeResourceBeforeDelta(NodeResource after)
             builder.setAspectNames(aspectsBefore);
         }
 
+        if (nodeTypeBefore != null)
+        {
+            builder.setNodeType(nodeTypeBefore.toPrefixString());
```
jamalkm, line 279: `NodeResourceHelper` class has a private method called `toString(QName k)` which uses the namespace service to get the prefix string. It makes sense to rename that method (e.g. `getQNamePrefixString`) and make it public so you can use it here instead of `nodeTypeBefore.toPrefixString()`.

SaraAspery: Done. Thanks Jamal :)

PR #991, opened 2020-05-14
REPO-5117: Node move event - primary parent only

cmshields:


src/main/java/org/alfresco/repo/event2/NodeResourceHelper.java
```diff
@@ -28,9 +28,12 @@
 import java.io.Serializable;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.Arrays;
```
jamalkm, line 32: Minor: unused imports.

cmshields: Imports removed.

PR #1005, opened 2020-05-21
Feature/repo 5118 node download event

SaraAspery:


src/main/java/org/alfresco/repo/node/AbstractNodeServiceImpl.java
```diff
@@ -1,28 +1,28 @@
-/*
```
jamalkm, line 1: There are a lot of whitespace changes in this file, can't we just revert it back to its original format?

.travis.yml
```diff
@@ -18,6 +18,7 @@ branches:
   only:
     - master
     - /support\/.*/
+    - feature/REPO-5118_Node_Download_Event
```
jamalkm, line 21: Please remove all the temp Dev release stuff.

src/main/java/org/alfresco/repo/node/AbstractNodeServiceImpl.java
cmshields: I have reverted the whitespace changes in this file.

.travis.yml
cmshields: Dev release stuff has been removed.

PR #1011, opened 2020-05-22
ATS-731: Add missing legacy pipelines to the local basePipelines json file

aymanthefirst:


src/main/resources/alfresco/transforms/0100-basePipelines.json
```diff
@@ -210,6 +252,17 @@
         "pdfRendererOptions",
         "imageMagickOptions"
       ]
+    },
+    {
+      "transformerName": "OutlookMsg2Office",
```
alandavis, line 257: Please follow the name pattern invented by Erik. In this case it would be "outlookMsgToLibreofficeViaText"

montgolfiere: cc @eknizat

KikoUK:
> Are there any existing tests that exercise the base pipeline ?
> 
> If so, do we need to update. If not, at least add sanity check to manual TestRail ?

The only test we've done so far is to disable legacy and transform service and ensure that the added pipelines are available in the local transform service and that worked. 

Since there is no way of requesting a particular transform through repo "manually", I've added a unit tests creation task to this ticket for the newly added pipelines.  ATS-763, see "Functionality to be added to repo tests" section, that is to cover the new pipelines added to the basePipelines json.

src/main/resources/alfresco/transforms/0100-basePipelines.json
aymanthefirst: This is has been changed now

src/main/resources/alfresco/transforms/0100-basePipelines.json
```diff
@@ -198,6 +198,48 @@
...
+    {
+      "transformerName": "libreofficeToPdfBoxViaPdf",
+      "transformerPipeline" : [
+        {"transformerName": "libreofficeToPdf", "targetMediaType": "application/pdf"},
+        {"transformerName": "PdfBox"}
+      ],
+      "supportedSourceAndTargetList": [
+      ],
+      "transformOptions": [
+      ]
```
PunchedChimera, line 241: Should these have options? "pdfboxOptions" are valid for the "PdfBox" transformer

src/main/resources/alfresco/transforms/0100-basePipelines.json
```diff
@@ -210,6 +252,17 @@
...
+    {
+      "transformerName": "outlookMsgToLibreofficeViaText",
+      "transformerPipeline" : [
+        {"transformerName": "OutlookMsg",                              "targetMediaType": "text/plain"},
+        {"transformerName": "libreoffice"}
+      ],
+      "supportedSourceAndTargetList": [
+      ],
+      "transformOptions": [
+      ]
```
PunchedChimera, line 265: Should these have options? "tikaOptions" are valid for the "OutlookMsg" transformer

montgolfiere:
Is this ready to merge ? We also need to copy over to ATS (new T-Router branch) ?

KikoUK:
> Is this ready to merge (cc @KikoUK, @eknizat @PunchedChimera) ?
> 
> We also need to copy over to ATS (new T-Router branch).
> 
> Does the existing base pipeline (prior to this change) have any sanity tests in Repo ? If yes, can we add one or two (or raise Test Automation jira). Alternatively, I assume it will be tested via ATS as above.

Yes, after a short investigation, I have discovered that the LocalRenditionTests use all available MimeTypes in this data-model [map](https://github.com/Alfresco/alfresco-data-model/blob/master/src/main/java/org/alfresco/repo/content/MimetypeMap.java) in this [loop](https://github.com/Alfresco/alfresco-repository/blob/ee9af641925be1c162f6cf2839cea81641ecd5fd/src/test/java/org/alfresco/repo/rendition2/AbstractRenditionTest.java#L225) as source and runs them against all targetMimeTypes in the same list in [here](https://github.com/Alfresco/alfresco-repository/blob/ee9af641925be1c162f6cf2839cea81641ecd5fd/src/test/java/org/alfresco/repo/rendition2/AbstractRenditionTest.java#L303). This mimeTypeMap file contains at least 2 steps of each pipeline we added to repo basePipeline json.

PR #1012, opened 2020-05-22
Added CMIS integration test for node downloaded event

adina-ababei:


src/test/java/org/alfresco/repo/event2/DownloadRepoEventIT.java
```diff
@@ -0,0 +1,196 @@
...
+import org.alfresco.service.cmr.repository.NodeRef;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * @author Adina Ababei
+ */
+public class DownloadRepoEventIT extends AbstractContextAwareRepoEvent
+{
+    ContentService contentService;
```
jamalkm, line 46: Is there a reason for the access modifier to be `package` rather than `private`?

adina-ababei: No, I changed it to `private`

PR #1031, opened 2020-06-03
MNT-21487: Switch from smaller schemas jar (poi-ooxml-schemas) to the full jar (ooxml-schemas)

amohammedalfresco:
- Replace `poi-ooxml-schemas` with  `ooxml-schemas`


pom.xml
```diff
@@ -527,6 +527,17 @@
...
+                <exclusion>
+                    <groupId>org.apache.poi</groupId>
+                    <artifactId>poi-ooxml-schemas</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.poi</groupId>
+            <artifactId>ooxml-schemas</artifactId>
+            <version>1.4</version>
```
alandavis, line 540: I note we updated the version of POI in https://github.com/Alfresco/alfresco-repository/pull/644/files recently. Should this be the same version? If nothing else the 1.4 should be defined in a <dependency.xxx.version>.

The other thing to check is if the inclusion introduces any vulnerabilities (check WhiteSource).

amohammedalfresco: I believe the latest poi and ooxml-schemas versions are good to be used but unfortunately the version convention is quite different for them.  I've defined `ooxml-schemas` dependency variable.  The WhiteSource scan reported no issues/voilations and passed as expected (report attached to the ticket).

PR #1046, opened 2020-06-12
ACS-306: ACS Repository: V1 Rest API for content direct access urls

cristianturlica:
- added alfresco.content.directAccessUrl.expiryTime.seconds property to allow the configuration of the direct access url expiry time
   - changed getDirectAccessUrl method signature to allow the successful execution of the acl interceptor
   - moved get content data to it's own method so it can be easily reused
   - updated test

### DIFF
--- a/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
+++ b/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
@@ -163,8 +163,8 @@ public interface ContentService extends ContentTransformService
      * @param expiryTime
      *            the expiration time in milliseconds of the direct access url. This
      *            is the length of time in milliseconds that the link is valid for.
-     * @return A direct access URL for a binary content or empty string if not
-     *         supported
+     * @return A direct access URL for a binary content or returns null if there is
+     *         no binary content for the node or empty string if not supported
      */
     @Auditable(parameters = {"nodeRef", "expiryTime"})
     public String getDirectAccessUrl(NodeRef nodeRef, int expiryTime);
--- a/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
+++ b/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
@@ -157,9 +157,10 @@ public interface ContentService extends ContentTransformService
      * Gets a presigned URL to directly access a binary content. It is up to the content store
      * if it can fulfil this request with an expiry time or not.
      *
-     * @param contentUrl A content store URL
+     * @param nodeRef a reference to a node having a content property
      * @param expiryTime Expiration time in milliseconds 
      * @return A direct access URL for a binary content or empty string if not supported
      */
-    public String getDirectAccessUrl(String contentUrl, int expiryTime);
+    @Auditable(parameters = {"nodeRef", "expiryTime"})
+    public String getDirectAccessUrl(NodeRef nodeRef, int expiryTime);
 }
--- a/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
+++ b/src/main/java/org/alfresco/service/cmr/repository/ContentService.java
@@ -154,12 +154,17 @@ public interface ContentService extends ContentTransformService
     public ContentWriter getTempWriter();
 
     /**
-     * Gets a presigned URL to directly access a binary content. It is up to the content store
-     * if it can fulfil this request with an expiry time or not.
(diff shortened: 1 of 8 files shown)


src/main/resources/alfresco/repository.properties
```diff
@@ -1327,4 +1327,8 @@ system.delete_not_exists.batchsize=100000
 system.delete_not_exists.delete_batchsize=1000
 system.delete_not_exists.read_only=false
 system.delete_not_exists.timeout_seconds=-1
-system.prop_table_cleaner.algorithm=V2
\ No newline at end of file
+system.prop_table_cleaner.algorithm=V2
+
+# Configure the expiration time of the direct access url.
+# Note: It is up to the actual ContentStore implementation if it can fulfil this request or not.
+alfresco.content.directAccessUrl.expiryTime.seconds=-1
```
ancutam, line 1334: What does `-1` mean? No expiry time?

src/main/java/org/alfresco/service/cmr/repository/ContentService.java
```diff
@@ -157,9 +157,10 @@ public ContentWriter getWriter(NodeRef nodeRef, QName propertyQName, boolean upd
      * Gets a presigned URL to directly access a binary content. It is up to the content store
      * if it can fulfil this request with an expiry time or not.
      *
-     * @param contentUrl A content store URL
+     * @param nodeRef a reference to a node having a content property
      * @param expiryTime Expiration time in milliseconds
```
gavincornwell, line 161: I think we need to be a little clearer with the explanation of the expiryTime parameter. Is it the time in epoch seconds when the link expires or the length of time in milliseconds that the link is valid for?

src/main/resources/alfresco/repository.properties
gavincornwell: I don't think it should be possible to get a link that never expires (not even sure the AWS API allows for that) so we should have a reasonable default i.e. 24 hours?

src/main/java/org/alfresco/service/cmr/repository/ContentService.java
cristianturlica: The length of time in milliseconds that the link is valid for. I will update the description.

src/main/resources/alfresco/repository.properties
cristianturlica: The initial intent was that -1 - would imply not set and allowed for each connecter default to be used. But I can change it to 24 hours.

cristianturlica: Amazon sdk has something similar and defaults to 15 minutes if not set...

cristianturlica: Set to 86400 seconds (24 hours)

src/main/java/org/alfresco/service/cmr/repository/ContentService.java
```diff
@@ -154,12 +154,18 @@ public ContentWriter getWriter(NodeRef nodeRef, QName propertyQName, boolean upd
...
-     * @param contentUrl A content store URL
-     * @param expiryTime Expiration time in milliseconds 
-     * @return A direct access URL for a binary content or empty string if not supported
+     * @param nodeRef
+     *            a reference to a node having a content property
+     * @param expiryTime
+     *            the expiration time in milliseconds of the direct access url. This
+     *            is the length of time in milliseconds that the link is valid for.
+     * @return A direct access URL for a binary content or empty string if not
+     *         supported
```
ancutam, line 167: or returns `null` if there is no binary content for the node

DenisGabriela: The AWS API generates an URL even though there is no binary content on the given path.

ancutam: Our ContentService returns null though.

DenisGabriela: You're right. I was thinking of binary content in s3.

src/main/resources/alfresco/repository.properties
cristianturlica: Updated to 28800 (8 hours) - after talking with Alex Lee

src/main/java/org/alfresco/service/cmr/repository/ContentService.java
cristianturlica: Change done, good catch.

PR #1047, opened 2020-06-12
Feature/repo 5204 revert download event

cmshields:


src/main/java/org/alfresco/opencmis/CMISConnector.java
```diff
@@ -1824,6 +1815,8 @@ public Event prepareEvent(String user, String networkId, String transactionId)
                 }
             }
         });
+
+
```
SaraAspery, line 1819: extra blank line

PR #1015, opened 2020-05-26
REPO-4334 Move metadata extraction into T-Engines

alandavis:
- ContentMetadataExtracter (the Action service Executer):
  - Code refactored so that post processing of extracted properties can be called after a
    transform has taken place in an async call. Previously done in the same Thread.
  - The source size is passed to the metadataExtracterRegistry as this is needed to select
    the transform.
  - Passes the source nodeRef to the metadataExtractor, so that it is possible to update
    the correct node after the async transform call.
- AbstractMappingMetadataExtracter (the base class for all extractors):
  - Post extract processing extracted into separate methods.
  - NodeRef made available to sub classes.
- AsynchronousExtractor (new metadata extractor and embedder) that will use
  a T-Engine.
- Set targetExtension in transform for metadata extract and embed requests to T-Engines
   in both LocalTransforms (community) and RenditionEventProducer (enterprise).
- RenditionService2Impl (used to request and respond to async transform requests).
  - Add consumeExtractedMetadata and consumeEmbeddedMetadata methods, which
    handle the response, normally be calling AsynchronousExtractor.
- Discovered that the JodConverterMetadataExtracter was superseded by the
  TikaPoweredMetadataExtracter for all the types that it support in 6.0.1 and could have been
  deprecated at that time as part of REPO-2910. It will now be deprecated so that it can be
  removed in 7.0.0.

Additional tests are being added to AsynchronousExtractorTest for overwritePolicy, enableStringTagging and carryAspectProperties in a separate PR.

PR #1051, opened 2020-06-17
REPO-4334 ThumbnailService needs its own transaction in RetryingTransactionHelper to avoid invalidating parent transaction

alandavis:
ThumbnailServiceImpl change reviewed by Stefan, while being written.

Branched from feature/REPO-4334_metadata hence all the other files. These were reviewed as part of https://github.com/Alfresco/alfresco-repository/pull/1015/ which has been merged and reverted, because of a failure found in acs-packaging.


alandavis:
It is a bug, but there is too much fall out from code that has tried to cope with the bug.

PR #1053, opened 2020-06-19
WIP: REPO-5212 Fix async metadata fallout.

alandavis:
As a result of making metadata extractors async #1015, we encountered CMIS TAS test failures in packaging projects related to failing to create thumbnails when uploading new content.

It has been possible to fix these by making the code that creates thumbnails for CMIS async too. This had already been done for Share and V1 uploads, but CMIS appears to have been missed.

However, this has had a knock on affect on tests for the deprecated ThumbnailService.

When reviewing this commit, please only look at the changes NOT in #1054 . A file list will be added prior to review as the change list currently includes files from that PR.


alandavis:
Closing as code has been moved to new projcets

PR #1081, opened 2020-07-06
REPO-5235: Use unforked tika

eknizat:
* Use latest unforked Tika version and exclude the forked version from alfresco-model.
* Bring back the old fix done in https://fisheye.alfresco.com/changelog/alfresco-enterprise?cs=103763

### DIFF
--- a/src/main/java/org/alfresco/repo/content/metadata/TikaAutoMetadataExtracter.java
+++ b/src/main/java/org/alfresco/repo/content/metadata/TikaAutoMetadataExtracter.java
@@ -70,6 +70,8 @@ public class TikaAutoMetadataExtracter extends TikaPoweredMetadataExtracter
     private static TikaConfig config;
     private static String EXIF_IMAGE_HEIGHT_TAG = "Exif Image Height";
     private static String EXIF_IMAGE_WIDTH_TAG = "Exif Image Width";
+    private static String NEW_EXIF_IMAGE_HEIGHT_TAG = "Exif SubIFD:Exif Image Height";
+    private static String NEW_EXIF_IMAGE_WIDTH_TAG = "Exif SubIFD:Exif Image Width";
     private static String JPEG_IMAGE_HEIGHT_TAG = "Image Height";
     private static String JPEG_IMAGE_WIDTH_TAG = "Image Width";
     private static String COMPRESSION_TAG = "Compression";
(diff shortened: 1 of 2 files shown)


montgolfiere:
I assume this means we can also resolve MNT-19477 (cc @ancutam ) ?

ancutam:
@eknizat @montgolfiere I was looking at https://issues.alfresco.com/jira/browse/REPO-1293, which mentions an attempt to revert a tika related patch in repo (https://issues.alfresco.com/jira/browse/MNT-13920), which didn't happen because of another bug (https://issues.apache.org/jira/browse/TIKA-2630).
Did you try a _full revert_ of changes done in MNT-13920 (the old fix mentioned in the PR description). The issue is covered by tests so it should be _easy_ to see if it fails without the initial fix.

src/main/java/org/alfresco/repo/content/metadata/TikaAutoMetadataExtracter.java
```diff
@@ -68,8 +68,8 @@
     protected static Log logger = LogFactory.getLog(TikaAutoMetadataExtracter.class);
     private static AutoDetectParser parser;
     private static TikaConfig config;
-    private static String EXIF_IMAGE_HEIGHT_TAG = "Exif Image Height";
-    private static String EXIF_IMAGE_WIDTH_TAG = "Exif Image Width";
+    private static String EXIF_IMAGE_HEIGHT_TAG = "Exif SubIFD:Exif Image Height";
+    private static String EXIF_IMAGE_WIDTH_TAG = "Exif SubIFD:Exif Image Width";
     private static String JPEG_IMAGE_HEIGHT_TAG = "Image Height";
     private static String JPEG_IMAGE_WIDTH_TAG = "Image Width";
     private static String COMPRESSION_TAG = "Compression";
```
ancutam, line 75: COMPRESSION_TAG is not used anymore.

eknizat:
> @eknizat @montgolfiere I was looking at https://issues.alfresco.com/jira/browse/REPO-1293, which mentions an attempt to revert a tika related patch in repo (https://issues.alfresco.com/jira/browse/MNT-13920), which didn't happen because of another bug (https://issues.apache.org/jira/browse/TIKA-2630).
> Did you try a _full revert_ of changes done in MNT-13920 (the old fix mentioned in the PR description). The issue is covered by tests so it should be _easy_ to see if it fails without the initial fix.

Yes, some test fails if the TikaAutoMetadataExtracter.extractSpecific() method is not overridden, in which case the "tiff:ImageLength" and "tiff:ImageWidth " values are equivalent to the "Exif..." values. From what I can tell in that test, they should be equivalent to the "Image Height" and "Image Width" instead.

eknizat:
> LGTM.
> Did we double-check that we do not have any changes in our fork on
> https://github.com/Alfresco/alfresco-tika
> that are not yet part of the official Tika?

Seems that all changes are in according to https://github.com/apache/tika/compare/master...Alfresco:master

PR #1101, opened 2020-07-15
REPO-5094: Allow switching between existing content store subsystems

ancutam:
* Update `CryptodocSwitchableApplicationContextFactory` to accept switching (via JMX) between existing content store subsystems, not just based on plain bean names.
* Added `EncryptedContentStoreChildApplicationContextFactory` as an extension of `ChildApplicationContextFactory` with support to specify if the content store in the subsystem is encrypted or not.

### DIFF
--- a/src/main/java/org/alfresco/repo/management/subsystems/CryptodocSwitchableApplicationContextFactory.java
+++ b/src/main/java/org/alfresco/repo/management/subsystems/CryptodocSwitchableApplicationContextFactory.java
@@ -136,7 +136,7 @@ public class CryptodocSwitchableApplicationContextFactory extends SwitchableAppl
 
     private boolean isEncryptionSupported()
     {
-        boolean isSupported = false;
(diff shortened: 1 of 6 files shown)


src/main/java/org/alfresco/repo/management/subsystems/CryptodocSwitchableApplicationContextFactory.java
```diff
@@ -128,23 +65,80 @@ public void setProperty(String name, String value)
         {
             if (!isUpdateable(name))
```
skopf, line 66: Why do we need this test? I cannot see that `isUpdateable` is overwritten anywhere, so it should return the default `true`.
And even if it is overwritten, I think the message on the Exception is wrong.

ancutam: True, I removed the test.

PR #1111, opened 2020-07-20
SEARCH-2273: Add detailed message when ignoring nodes in SOLR without…

aborroy:
… type or with types not registered in Dictionary Service.

### DIFF
--- a/src/main/java/org/alfresco/repo/solr/SOLRTrackingComponentImpl.java
+++ b/src/main/java/org/alfresco/repo/solr/SOLRTrackingComponentImpl.java
@@ -873,11 +873,7 @@ public class SOLRTrackingComponentImpl implements SOLRTrackingComponent
                     typeQName = nodeDAO.getNodeType(nodeId);
                     if (typeQName != null)
                     {
-                        typeDefinition = dictionaryService.getType(typeQName);
-                        if (typeDefinition == null)
-                        {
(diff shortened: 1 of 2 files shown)


src/main/java/org/alfresco/repo/solr/SOLRTrackingComponentImpl.java
```diff
@@ -865,7 +865,26 @@ public void getNodesMetadata(
...
+                    String errorMessage = "NodeId " + nodeId + " with nodeRef " + nodeRef;
+
+                    typeQName = nodeDAO.getNodeType(nodeId);
+                    if (typeQName != null)
+                    {
+                        typeDefinition = dictionaryService.getType(typeQName);
+                        if (typeDefinition == null)
+                        {
+                            errorMessage += " has type " + typeQName + ", but this type is not registered in DictionaryService.";
+                        }
```
tpage-alfresco, line 880: Missing an "else" here.

aborroy: Thanks, I've simplified the code at this point.

tpage-alfresco:
This looks fine to me, but I'm a bit concerned by the amount of code added to test it.  Instead we could extract the new production code into a (non-private) method and call it directly from a unit test.  I think that unit test would be about 10 lines of code.  If you can see more value in testing this with an IT then we can leave it though.

PR #1115, opened 2020-07-21
MNT-21317 Move on Update Folder Rule triggers unexpectedly for MOVED content.

nithinnambiar:
This change is to fix the issue related to https://issues.alfresco.com/jira/browse/MNT-21317.  This would ignore 'cascadeCRC' property updates inside OnPropertyUpdateRuleTrigger when it checks for any changes to properties before triggering the rules.


nithinnambiar:
> Does the test fail without the fix?
> It is not clear how ContentModel.PROP_CASCADE_CRC can affect the test results as it is not used in the test.

Yes, the rule gets triggered twice without the fix  as the move updates ContentModel.PROP_CASCADE_CRC.

killerboot:
Is it possible to add a test that is using the ContentModel.PROP_CASCADE_CRC property directly?

nithinnambiar:
> Is it possible to add a test that is using the ContentModel.PROP_CASCADE_CRC property directly?

I have refactored the test now to update ContentModel.PROP_CASCADE_CRC directly in the test.

PR #1149, opened 2020-08-04
MNT-21614: LDAP sync communication error leads to group members deletion

amohammedalfresco:
- Catch reported exceptions to filter ldap sync aborts


src/main/java/org/alfresco/repo/security/sync/ldap/LDAPUserRegistry.java
```diff
@@ -882,6 +883,26 @@ else if (hasAttributeValue(objectClass, LDAPUserRegistry.this.groupType))
                                             continue;
                                         }
                                     }
+                                    catch (ServiceUnavailableException e)
```
killerboot, line 886: Both exceptions have the same outcome, please join them in one catch.

amohammedalfresco: Sure.  Thanks.

PR #1189, opened 2020-09-04
REPO-5339 Improve concurrency of SimplePermissionReference

killerboot:
The piece of code before changes seemed to be designed to provide a thread safe factory for singletons. I didn't find any particular reason for this choice by looking at the usages of SimplePermissionReference in the code. It actually may be sufficient to just create new SimplePermissionReference objects each time.

### DIFF
--- a/src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
+++ b/src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
@@ -25,7 +25,7 @@
  */
 package org.alfresco.repo.security.permissions.impl;
 
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.HashMap;
 
 import org.alfresco.service.namespace.QName;
 import org.alfresco.util.Pair;
@@ -39,7 +39,7 @@ public final class SimplePermissionReference extends AbstractPermissionReference
 {   
     private static final long serialVersionUID = 637302438293417818L;
 
-    private static ConcurrentHashMap<Pair<QName, String>, SimplePermissionReference> instances = new ConcurrentHashMap<>();
+    private static HashMap<Pair<QName, String>, SimplePermissionReference> instances = new HashMap<>();
 
     /**
      * Factory method to create simple permission references
@@ -52,9 +52,15 @@ public final class SimplePermissionReference extends AbstractPermissionReference
         SimplePermissionReference instance = instances.get(key);
         if (instance == null)
         {
-            // TODO This may result in duplicate objects, but it is most likely does not matter
-            instance = new SimplePermissionReference(qName, name);
-            instances.put(key, instance);
+            synchronized (SimplePermissionReference.class)
+            {
+                instance = instances.get(key);
+                if (instance == null)
+                {
+                    instance = new SimplePermissionReference(qName, name);
+                    instances.put(key, instance);
+                }
+            }
         }
         return instance;
     }
--- a/src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
+++ b/src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
@@ -25,15 +25,9 @@
  */
 package org.alfresco.repo.security.permissions.impl;
 
-import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
```diff
@@ -40,59 +39,32 @@
 {   
     private static final long serialVersionUID = 637302438293417818L;
 
-    private static ReadWriteLock lock = new ReentrantReadWriteLock();
-
-    private static HashMap<QName, HashMap<String, SimplePermissionReference>> instances = new HashMap<QName, HashMap<String, SimplePermissionReference>>();
+    private static HashMap<Pair<QName, String>, SimplePermissionReference> instances = new HashMap<>();
```
nithinnambiar, line 42: Just an observation. Wouldn't it be better to replace the HashMap with a ConcurrentHashMap? That could help in avoiding locks and improve scalability.

killerboot: As far as I understood the ConcurrentHashMap provides a copy for all readers while the write operation happens. This means that the change like this https://github.com/Alfresco/alfresco-repository/pull/1189/commits/92c29c01615f7828cebc089904d722761bc8caf2 may result in duplicate objects if parallel threads got null and updated the entry one after another. On the other hand if the write operation is fully synchronised, then there is no reason for ConcurrentHashMap in this case.

nithinnambiar: Even though it doesn't lock on Read operations the Map could be updated atomically using computeIfAbsent function to avoid any duplicates.

skopf (changes requested):
Can we really update a HashMap while concurrently reading from it?

src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
```diff
@@ -1,98 +1,99 @@
...
      * 
      * @return a simple permission reference
      */
-    public static SimplePermissionReference getPermissionReference(QName qName, String name)
-    {
-        lock.readLock().lock();
-        try
+    public static SimplePermissionReference getPermissionReference(QName qName, String name) {
+        //loop if thread is interrupted and should be run again
+        while (true)
```
nithinnambiar, line 58: Just wondering if the method could be simplified like this.

` return instances.computeIfAbsent(key, pair -> new SimplePermissionReference(pair.getFirst(), pair.getSecond()));`

skopf (changes requested):
I checked the source of ConcurrentHashMap. We can actually read from this map without blocking. So this would be faster than the current implementation.

But I don't get why we need to create a FutureTask for the SimplePermissionReference and why this ConcurrentHashMap actually needs to hold factories for SimplePermissionReference. Why can't we simply put these elements in the map directly?

In the case that there is no element in the map yet, it could happen that a thread creates a new object but then doesn't need it because the putIfAbsent() indicates that some other thread has actually put it in the map meanwhile.
But we still have the same problem. We just shifted this from SimplePermissionReference to FutureTask

jottley:
@skopf This a good explanation of the reasons for using Future in "Java Concurrency in Practice". It is a long (6 pages with code examples) explanation so I'll send the pages to you by email.  The tl;dr of it is there still is a small window which we would try to reduce the chance of two threads trying to insert the same values.  The example from the book is actually computing a value so that window is in the compute...we don't have that so it could be reasonable to remove the Future and just go with a straight concurrentmap.

src/main/java/org/alfresco/repo/security/permissions/impl/SimplePermissionReference.java
jottley: This actually ends up being slightly slower though I do agree that it is simpler

nithinnambiar: I don't think it is going to be any slower since the computation inside the lambda isn't complex. Also `computeIfAbsent` would avoid unnecessary creation of SimplePermissionReference objects when multiple threads are acting on the same key.

skopf (approved):

PR #1205, opened 2020-09-15
WIP: REPO-5325 Determine if search accelerated tables support query

alandavis:
- Start moving code on alfresco-repository feature/REPO-5325_TMDQ_Accelerated_Table_Support_Detection to alfresco-enterprise-repo.

Nana, Bruno, this is just an initial first step at moving your code to alfresco-enterprise-repo. The idea is to give you an idea of where to place stuff. You will see there is a new EnterpriseDBQuery class that extends DBQuery. The QuerySet has been moved. Clearly it will not compile as there are lots of other changes in your code.

The real work is for you to work out how to get the <select id="select_byDynamicQuery"... defined in enterprise to be picked up when running the enterprise edition. It should be possible as there are lots of other examples in under alfresco-enterprise-repo src/main/resources/alfresco/ibatis.

Can you also try to keep the style of the code the same as it is at the moment. The "Kernighan and Ritchie" indentation style is not used by Alfresco. We always use { } even for simple statements. There is also a space between if and (. Mario is wanting to introduce more formal rules, so I will not dig out our current repo rules, but making it look the same is a good idea regardless of what happens next.

Also see https://github.com/Alfresco/alfresco-enterprise-repository/pull/476 for the enterprise changes.


bbossola:
> Nana, Bruno, this is just an initial first step at moving your code to alfresco-enterprise-repo. The idea is to give you an idea of where to place stuff. You will see there is a new EnterpriseDBQuery class that extends DBQuery. The QuerySet has been moved. Clearly it will not compile as there are lots of other changes in your code.

As I mentioned when we spoke earlier, and as you probably saw, this enhancement requires several changes in different areas of the codebase and, unfortunately, I am afraid cannot be solved by just extending a few classes. However, we will follow this direction and do our best find a suitable solution so that the code firmly lays in the enterprise repository only,

 
> The real work is for you to work out how to get the <select id="select_byDynamicQuery"... defined in Enterprise to be picked up when running the enterprise edition. It should be possible as there are lots of other examples in under alfresco-enterprise-repo src/main/resources/alfresco/ibatis.

Thanks, we will check those out.

 
> Can you also try to keep the style of the code the same as it is at the moment. The "Kernighan and Ritchie" indentation style is not used by Alfresco. We always use { } even for simple statements. There is also a space between if and (. Mario is wanting to introduce more formal rules, so I will not dig out our current repo rules, but making it look the same is a good idea regardless of what happens next.

Actually we did our best to write the new code following a convention that was similar to the one used in the existing code, If we were to use modern conventions you can be sure the code would have been really different :) I am not sure I fully grasp the reference to the K&R notation (even having my copy from my first C job in 1988 handy) but let me see if I understand what you meant:

1. _"there is also a space between if and ("_
never put a space between a control instruction (if/while...) and the expression. Expected:
```
if(x==2)
{
    ....
}
```
2. _"always use { } even for simple statements"_
always use brackets to define a block even for single line statement when used in a control statement. Expected:
```
if(x==2)
{
    System.out.println("Hello");
}
```

Assuming this is the summary is correct, we are happy to enforce those rules in our code formatted, as long as we can apply then across the code. If you have full detailed code conventions, these are definitely welcome.

Please be also aware that In regards to (1), you can find occurrences of the space-after-if on master in existing code, such as in the [same class](https://github.com/Alfresco/alfresco-repository/blob/09ecef9e4209e7e626c301353e91560eaf6f3564/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/db/DBQueryEngine.java#L151) and in others, like in [LuceneQueryEngine](https://github.com/Alfresco/alfresco-repository/blob/09ecef9e4209e7e626c301353e91560eaf6f3564/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/lucene/LuceneQueryEngine.java#L175) and [CalendarServiceImpl](https://github.com/Alfresco/alfresco-repository/blob/09ecef9e4209e7e626c301353e91560eaf6f3564/src/main/java/org/alfresco/repo/calendar/CalendarServiceImpl.java#L162) (I picked those two randomly).

In regards to (2), I believe you refer to this [single occurence](https://github.com/Alfresco/alfresco-repository/blob/1158cb6d1f2fd52b0293c4a021f64c185c9f5ab3/src/main/java/org/alfresco/repo/search/impl/querymodel/impl/db/DBQuery.java#L769) across the 754 changes. Is that correct, or did we accidentally missed this in other parts? 

Please let us know if you want us to aggressively use the formatter to follow the conventions you mentioned in every class we touch.

alandavis:
This code has been moved to the alfresco-community-repo project.

